PR #14, opened 2024-03-25
Add static test scripts

Rozzii:
In order to prepare the repository for CI integration

This commit:
- Adds a subset of the static analysis and formatting tools.
- Modifies the root Makefile to allow running gomod.sh
- aligns static test container wrapper scripts

I will align the code and the docs in multiple PRs to not create xK line PRs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.21.7@sha256:549dd88a1a53715f177b41ab5fee25f7a376a6bb5322ac7abe263480d9554021
+ARG BUILD_IMAGE=docker.io/golang:1.21.8@sha256:856073656d1a517517792e6cdd2f7a5ef080d3ca2dff33e518c8412f140fdd2d
 ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
 
 # Build the manager binary
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,28 @@
+# Setting SHELL to bash allows bash commands to be executed by recipes.
+# Options are set to exit when a recipe line exits non-zero or a piped command fails.
+SHELL = /usr/bin/env bash -o pipefail
+.SHELLFLAGS = -ec
+
+.DEFAULT_GOAL:=help
+
+GO_VERSION ?= 1.21.8
+GO := $(shell type -P go)
+# Use GOPROXY environment variable if set
+GOPROXY := $(shell $(GO) env GOPROXY)
+ifeq ($(GOPROXY),)
+GOPROXY := https://proxy.golang.org
+endif
+export GOPROXY
+
+# Drectories.
+# Full directory of where the Makefile resides
+ROOT_DIR := $(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+APIS_DIR := api
+TEST_DIR := test
+BIN_DIR := bin
+
+# Active module mode, as we use go modules to manage dependencies
+export GO111MODULE=on
 # VERSION defines the project version for the bundle.
 # Update this value when you upgrade the version of your project.
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
@@ -56,11 +81,6 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
-# Setting SHELL to bash allows bash commands to be executed by recipes.
-# Options are set to exit when a recipe line exits non-zero or a piped command fails.
-SHELL = /usr/bin/env bash -o pipefail
-.SHELLFLAGS = -ec
-
 RUNARGS ?=
 
 .PHONY: all
@@ -88,6 +108,15 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+
+.PHONY: modules
+modules: ## Runs go mod to ensure proper vendoring.
+	$(GO) mod tidy
+	$(GO) mod verify
+	cd $(APIS_DIR) && $(GO) mod tidy
+	cd $(APIS_DIR) && $(GO) mod verify
+	cd $(TEST_DIR) && $(GO) mod tidy
+	cd $(TEST_DIR) && $(GO) mod verify
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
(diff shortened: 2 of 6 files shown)


tuminoid:
> I will align the code and the docs in multiple PRs to not create xK line PRs.

Is there a lot of issues that would fail a test? Otherwise I'd suggest enabling Prow first, so each PR could be tested.

Makefile
```diff
@@ -59,8 +84,8 @@ GOBIN=$(shell go env GOBIN)
 endif
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
-# Options are set to exit when a recipe line exits non-zero or a piped command fails.
-SHELL = /usr/bin/env bash -o pipefail
+# Optiona are set to exit when a recipe line exits non-zero or a piped command fails.
```
dtantsur, line 87: Options

Makefile
```diff
@@ -69,6 +94,8 @@ RUNARGS ?=
 all: build
 
 ##@ General
+            self._logger.debug("boot image id is: " + str(image_id))
+            aelf._logger.debug("boot image id is: " + str(image_id))
```
dtantsur, line 98: Unrelated changes

Makefile
```diff
@@ -91,6 +118,17 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
+.PHONY: modules
+modules: ## Runs go mod to ensure proper vendoring.
+	$(GO) mod tidy
+	$(GO) mod verify
+	cd $(TOOLS_DIR) && $(GO) mod tidy
```
dtantsur, line 125: Note that apis and tools are not separate modules here (yet?)

go.mod
```diff
@@ -1,80 +1,105 @@
 module github.com/metal3-io/ironic-standalone-operator
 
-go 1.19
+go 1.22.0
```
dtantsur, line 3: Please don't go to 1.22.0, it will break local development for virtual everyone. The main branch has 1.21 now.

test/go.mod
```diff
@@ -1,37 +1,41 @@
 module github.com/metal3-io/ironic-standalone-operator/test
 
-go 1.19
+go 1.22.0
```
dtantsur, line 3: Same here

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.21.7@sha256:549dd88a1a53715f177b41ab5fee25f7a376a6bb5322ac7abe263480d9554021
+ARG BUILD_IMAGE=docker.io/golang:1.22@sha256:0b55ab82ac2a54a6f8f85ec8b943b9e470c39e32c109b766bbc1b801f3fa8d3b
```
tuminoid, line 2: Specify the patch version as well. `1.22` is moving tag.

Makefile
```diff
@@ -1,3 +1,27 @@
+SHELL:=/usr/bin/env bash
+
+.DEFAULT_GOAL:=help
+
+GO_VERSION ?= 1.21.8
```
tuminoid, line 5: This is also mismatching versus Dockerfile. This is the right version to go-to as Dmitry pointed out.

go.mod
tuminoid: Also going down to patch level is different from other repos, and will have effect on consumers. While it might be the right thing to do going down the line, we might need to agree on it and implement it the same way for all repos, IMO.

hack/gomod.sh
```diff
@@ -0,0 +1,44 @@
...
+        exit 1
+    fi
+
+else
+    "${CONTAINER_RUNTIME}" run --rm \
+        --env IS_CONTAINER=TRUE \
+        --volume "${PWD}:/workdir:ro,z" \
+        --entrypoint sh \
+        --workdir /workdir \
+        docker.io/golang:1.22 \
```
tuminoid, line 42: 1.22 here too

tuminoid (commented):
Some nits.

test/go.mod
```diff
@@ -1,37 +1,41 @@
...
+
+toolchain go1.22.1
```
elfosardo, line 5: please remember to change toolchain version too

dtantsur: I'd rather drop this `toolchain` instruction. I don't understand why Go is sometimes adding it, but it's not required.

Dockerfile
Rozzii: but I have added the sha, shouldn't that specify it explicitly?

tuminoid: It does, but doesn't mean anything to humans, and some tools like Quay build will happily explode when those won't match later on. Also any tooling reading that doesn't need to do detective work to figure out which tag would it be.

Makefile
Rozzii: done

Makefile
Rozzii: removed thanks

Makefile
Rozzii: I have removed the tool dir for the moment.

Makefile
```diff
@@ -58,7 +80,7 @@ endif
 
 # Setting SHELL to bash allows bash commands to be executed by recipes.
 # Options are set to exit when a recipe line exits non-zero or a piped command fails.
-SHELL = /usr/bin/env bash -o pipefail
+SHELa = /usr/bin/env bash -o pipefail
```
tuminoid, line 83: Some weird typo here...

```suggestion
SHELL = /usr/bin/env bash -o pipefail
```

tuminoid (changes requested):

Makefile
Rozzii: nice catch thanks, no clue how this happened.

hack/ensure-go.sh
```diff
@@ -0,0 +1,49 @@
...
+Can't find 'go' in PATH, please fix and retry.
+See http://golang.org/doc/install for installation instructions.
+EOF
+        return 2
+    fi
+
+    local go_version
+    IFS=" " read -ra go_version <<< "$(go version)"
+    local minimum_go_version
+    minimum_go_version=go1.20
```
tuminoid, line 35: Should this be 1.21?

hack/gomod.sh
```diff
@@ -0,0 +1,44 @@
...
+        exit 1
+    fi
+
+else
+    "${CONTAINER_RUNTIME}" run --rm \
+        --env IS_CONTAINER=TRUE \
+        --volume "${PWD}:/workdir:ro,z" \
+        --entrypoint sh \
+        --workdir /workdir \
+        docker.io/golang:1.21.8 \
```
tuminoid, line 42: For hack scripts we don't specify patch version, so it will use the latest of the minor.

```suggestion
        docker.io/golang:1.21 \
```

hack/ensure-go.sh
Rozzii: done

hack/gomod.sh
Rozzii: done

tuminoid (approved):
LGTM

Rozzii:
/override markdownlint 
markdownlint fixes are in a separate PR let's ignore this.

PR #42, opened 2024-06-05
Mass-update dependencies

dtantsur:
We're about to pick up active development here again, let's make sure
to work with the latest stuff to avoid accidentally depending on old and
incompatible versions (looking at you, controller-runtime).

Forces bumping the Go version to 1.22 (using 1.22.2 to avoid reverting
the fix for https://osv.dev/vulnerability/GO-2024-2687).


dtantsur:
/assign @tuminoid 

Could you make sure I'm not undoing any of your security fixes?

tuminoid:
> /assign @tuminoid
> 
> Could you make sure I'm not undoing any of your security fixes?

You should be bumping to Go 1.22.3, 1.22.2 has 3 vulnerabilities open. Otherwise the tree is clean. :+1:

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.21.10@sha256:16438a8e66c0c984f732e815ee5b7d715b8e33e81bac6d6a3750b1067744e7ca
+ARG BUILD_IMAGE=docker.io/golang:1.22.4@sha256:5eb6d52cd78951fa53884e208a030db506ab619a13157b9b7ea3a8eb5e4cbeee
```
tuminoid, line 2: This is 1.22.4

Makefile
```diff
@@ -5,7 +5,7 @@ SHELL = /usr/bin/env bash -o pipefail
 
 .DEFAULT_GOAL:=help
 
-GO_VERSION ?= 1.21.10
+GO_VERSION ?= 1.22.2
```
tuminoid, line 8: But this is 1.22.2

go.mod
```diff
@@ -1,34 +1,34 @@
 module github.com/metal3-io/ironic-standalone-operator
 
-go 1.21
+go 1.22.2
```
tuminoid, line 3: Same here.

tuminoid (commented):
At least Dockerfile and Makefile versions should be aligned?

go.mod
dtantsur: My thinking was: the version in the Dockerfile is the actually used version, while this is a minimum version. I don't have 1.22.4 locally yes, so I cannot even bump it. I can downgrade the Dockerfile, but it seems like a missed opportunity.

dtantsur: I do have 1.22.3 locally though, I can align on this version if you're fine with it.

tuminoid: This is mininum yes, and this is the only go.mod file right now where we'd have patch mentioned. 1.22.2 is fine, but if you can make it plain 1.22, that would be in line with others.

Makefile
tuminoid: This is used by many GH actions to setup correct go version in workflows etc. It is a helper, and should match Dockerfile.

go.mod
tuminoid: OK, it seems it will not accept plain 1.22, but will replace it with 1.22.0 and insert toolchain, if its under 1.22.2.

Let's do 1.22.3 in go.mod and 1.22.4 in dockerfile/makefile?

dtantsur:
@tuminoid done as you suggested

dtantsur:
@tuminoid looks like github actions only have 1.22.2 on their ubuntu runners :(

tuminoid:
> @tuminoid looks like github actions only have 1.22.2 on their ubuntu runners :(

Yeah, let me take a look, if we can fix that. We have other repos working fine, so I think we need to add the go-setup step here as well. 

We also need to merge this for the gomod to pass: https://github.com/metal3-io/project-infra/pull/783

tuminoid:
> > @tuminoid looks like github actions only have 1.22.2 on their ubuntu runners :(
> 
> Yeah, let me take a look, if we can fix that. We have other repos working fine, so I think we need to add the go-setup step here as well.
> 
> We also need to merge this for the gomod to pass: [metal3-io/project-infra#783](https://github.com/metal3-io/project-infra/pull/783)

OK, so if we merge #43 and then rebase this, and also merge the project-infra PR for the gomod test, then I think this should pass.

tuminoid:
Also edit `hack/gomod.sh` to run Go 1.22 (fixes local runs).

/test gomod

hack/gomod.sh
```diff
@@ -34,11 +34,12 @@ if [ "${IS_CONTAINER}" != "false" ]; then
...
 
 else
+    GO_VERSION=$(make go-version)
     "${CONTAINER_RUNTIME}" run --rm \
         --env IS_CONTAINER=TRUE \
         --volume "${PWD}:/workdir:ro,z" \
         --entrypoint sh \
         --workdir /workdir \
-        docker.io/golang:1.21 \
+        docker.io/golang:${GO_VERSION} \
```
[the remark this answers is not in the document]
dtantsur: Well, it's a bit annoying to hardcode different variations of the same thing everywhere. Should we maybe have a text file that will provide a version that the CI uses? Or somehow de-duplicate it?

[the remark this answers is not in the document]
tuminoid: Golang bump is the least annoying and least coupled of them all, since you need to bump it once per minor version of golang, ie. once every six months. 

Project-infra vs repository coupling is annoying I agree, as I've done multiple non-compatible bumps to other linters etc and it requires a lot of juggling and coordination. So far we've deemed it to happen rarely enough to use time and effort to replace it as its how the Prow testing works right now with a dozen repos structure we have in place.

tuminoid (commented):
AFAIK this looks like solid golang bump.

/lgtm

Maybe update title to reflect that as its more about golang bump than deps.

dtantsur:
It was about dependencies initially, Go is sort of forced on me :)

/approve

dtantsur:
> @tuminoid: adding LGTM is restricted to approvers and reviewers in OWNERS files.

Indeed: https://github.com/metal3-io/ironic-standalone-operator/pull/44

PR #43, opened 2024-06-10
:seedling: add scaffolding for go bump to 1.22

tuminoid:
- Add "make go-version" in Makefile to easily access used Go version
- Add setup-go steps to functional and build workflows that setup the correct version of Go
- Bump gomod prow test to 1.22 (pairs with PR in project-infra: https://github.com/metal3-io/project-infra/pull/783)

This should allow the actual bump to golang 1.22 to succeed in https://github.com/metal3-io/ironic-standalone-operator/pull/42


hack/gomod.sh
```diff
@@ -39,6 +39,6 @@ else
         --volume "${PWD}:/workdir:ro,z" \
         --entrypoint sh \
         --workdir /workdir \
-        docker.io/golang:1.21 \
+        docker.io/golang:1.22 \
```
tuminoid, line 42: This should be part of #42 actually, and not be here.

tuminoid: Removed.

PR #45, opened 2024-06-12
Determine whether initcontainer is needed through IPAConfig options

matthewei:
Some times we can build a http server out of cluster and we can also ues it to be as a image-server. And we can upload kernal and ramdisk into it. So I think we may don't need the init-Pod in Ironic's Pod.

I have modified the Ironic CRD

```yaml
apiVersion: metal3.io/v1alpha1
kind: Ironic
metadata:
  name: ironic-server
spec:
  images:
    ipaConfig:
      ipaEnabled: true
      ironic: registry.paas/metal3-io/ironic:v1.0
      ramdiskDownloader: registry.paas/metal3-io/ipa-config:v17
```

### DIFF
--- a/api/v1alpha1/ironic_types.go
+++ b/api/v1alpha1/ironic_types.go
@@ -144,6 +144,11 @@ type Images struct {
 	// +kubebuilder:default=quay.io/metal3-io/ironic-ipa-downloader
 	// +optional
 	RamdiskDownloader string `json:"ramdiskDownloader,omitempty"`
+
+	// DisableRamdiskDownloader turns off the ramdisk downloader.
+	// +kubebuilder:default=false
+	// +optional
+	DisableRamdiskDownloader bool `json:"disableRamdiskDownloader,omitempty"`
 }
 
 // IronicSpec defines the desired state of Ironic
--- a/api/v1alpha1/ironic_webhook.go
+++ b/api/v1alpha1/ironic_webhook.go
@@ -203,7 +203,7 @@ func validateIronic(ironic *IronicSpec, old *IronicSpec) error {
 		}
 	}
 
-	if ironic.Images.AgentDownloadURL != "" {
+	if !ironic.Images.DisableRamdiskDownloader && ironic.Images.AgentDownloadURL != "" {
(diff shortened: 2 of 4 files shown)


matthewei:
@dhellmann Hi , Could you have a look my code.

api/v1alpha1/ironic_types.go
```diff
@@ -124,7 +124,8 @@ type Networking struct {
 	MACAddresses []string `json:"macAddresses,omitempty"`
 }
 
-type Images struct {
+type IPAConfig struct {
```
dtantsur, line 127: I'd rather keep the Images structure intact, especially since what you're adding is not an "IPA config", nor are you "enabling IPA". IPA is pretty much required for Ironic, you're enabling/disabling its downloader.

Let's just add disableRamdiskDownloader as a top-level option? (disable instead of enable so that it can default to false).

matthewei: OK, I will modify my code.

matthewei: @dtantsur  Is it ok?

api/v1alpha1/ironic_types.go
```diff
@@ -144,6 +144,11 @@ type Images struct {
 	// +kubebuilder:default=quay.io/metal3-io/ironic-ipa-downloader
 	// +optional
 	RamdiskDownloader string `json:"ramdiskDownloader,omitempty"`
+
+	// DisableRamdiskDownloader turns off the ramdisk downloader.
+	// +kubebuilder:default=false
+	// +optional
+	DisableRamdiskDownloader bool `json:"disableRamdiskDownloader,omitempty"`
```
dtantsur, line 151: This should not be part of Images, especially since the Images structure in its current form will be gone.

matthewei: OK, I would like to refactor my code.

dtantsur:
Hi! Are you still interested in this change? I'll be doing some work around ramdisk settings and may as well include your change if you don't have time to update it.

dtantsur:
Implemented in https://github.com/metal3-io/ironic-standalone-operator/pull/84

PR #47, opened 2024-07-31
Rework handling of images and versions

dtantsur:
User-supplied images via CRD's goes against operator best practices and
make upgrades very complicated. Remove the Images structure but allow
for downstream customizations via CLI flags and environment variables.

Record the version that was installed to use it in upgrade logic.


pkg/ironic/version.go
```diff
@@ -0,0 +1,37 @@
+package ironic
+
+const (
+	installedVersion string = "latest"
```
elfosardo, line 4: wondering if we want to have a different name for this (e.g. ironicInstalledVersion) as we talked about releasing ipa-downloader, and in the future we may need another const for that

elfosardo (commented):
looks good
just a comment before adding LGTM

pkg/ironic/version.go
dtantsur: Possibly, but it's a private variable, we can deal with it later.

PR #50, opened 2024-08-27
:seedling: Add wf to automatically build controller image

mquhuy:
This PR adds support for automatic container image build. Whenever a new PR is merged into `main` or one of the release branches, and whenever a new `v*` tag is created, a new container image is built and pushed to `quay.io/metal3-io`, with suitable tags applied.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -1,0 +1,25 @@
+name: build-images-action
+
+permissions:
+  contents: read
+
+on:
+  push:
+    branches:
+    - 'main'
+    - 'release-*'
+    tags:
+    - 'v*'
+
+jobs:
+  build_controller:
+    name: Build Ironic-standalone-operator container image
+    if: github.repository == 'metal3-io/ironic-standalone-operator'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: 'ironic-standalone-operator'
+      pushImage: true
+    secrets:
+      QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+      QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+      SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}


.github/workflows/build-images-action.yml
```diff
@@ -0,0 +1,25 @@
...
+on:
+  push:
+    branches:
+    - 'main'
+    - 'release-*'
+    tags:
+    - 'v*'
+
+jobs:
+  build_controller:
```
tuminoid, line 15: Are we expecting to have many jobs here? If not, then I'd probably not use controller in naming, but operator, or ISO (like IPAM) as more descriptive.

.github/workflows/build-images-action.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Build Ironic-standalone-operator-controller container image
+    if: github.repository == 'metal3-io/ironic-standalone-operator'
+    uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main
+    with:
+      image-name: 'ironic-standalone-operator/controller'
```
tuminoid, line 20: Same here, and I'm not even sure we actually want to have it named like that, but just `ironic-standalone-operator`  or even `ironic-standalone-operator-controller`.

tuminoid (commented):
We don't postfix any other image with `controller` either. IMO we shouldn't do that here either.

.github/workflows/build-images-action.yml
mquhuy: I was following the naming convention in `Makefile`. Afaik there's at least bundle image, although the Dockerfile for it is nowhere to be found. But I agree `ironic-standalone-operator-controller` would be a good name.

@dtantsur Wdyt?

.github/workflows/build-images-action.yml
mquhuy: I think there might be more images coming, based on what I found on the Makefile. IMO `operator` is not more descriptive though, ISO might be, but it's not yet a familiar acronym yet, at least for me. `controller` is, therefore, more suitable, as anyone could understand what it builds from the name. This is anyway just the name of the job, not the image name.

tuminoid:
@dtantsur you asked for workflow to build the images, would you mind taking a look and weighting in on job naming?

.github/workflows/build-images-action.yml
dtantsur: Makefile is basically auto-generated. I'd definitely use just ironic-standalone-operator to be consistent with baremetal-operator.

Rozzii: @mquhuy If you change the name according to what @dtantsur mentioned above, I am fine with putting lgtm on it.

mquhuy: Sorry I missed the comment. Change coming up!

tuminoid (commented):
/lgtm

Rozzii (commented):
/approve

dtantsur:
/approve

Thanks!

PR #53, opened 2024-10-28
Update test/prepare.sh to work with podman

dtantsur:

### DIFF
--- a/test/prepare.sh
+++ b/test/prepare.sh
@@ -2,18 +2,39 @@
 
 set -eux -o pipefail
 
-IMG="${IMG:-controller:test}"
+IMG="${IMG:-localhost/controller:test}"
 LOGDIR="${LOGDIR:-/tmp/logs}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
 
 mkdir -p "${LOGDIR}"
 
+if [ -z "${CONTAINER_RUNTIME}" ]; then
+    if podman version > /dev/null 2>&1;  then
+        CONTAINER_RUNTIME=podman
+    else
+        CONTAINER_RUNTIME=docker
+    fi
+fi
+
+kind_load() {
+    local image="$1"
+    if [ "${CONTAINER_RUNTIME}" = podman ]; then
+        local archive="$(mktemp --suffix=.tar)"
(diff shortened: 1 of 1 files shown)


test/prepare.sh
```diff
@@ -2,18 +2,39 @@
...
 set -eux -o pipefail
 
-IMG="${IMG:-controller:test}"
+IMG="${IMG:-localhost/controller:test}"
 LOGDIR="${LOGDIR:-/tmp/logs}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
 
 mkdir -p "${LOGDIR}"
 
+if [ -z "${CONTAINER_RUNTIME}" ]; then
```
tuminoid, line 11:
```suggestion
if [[ -z "${CONTAINER_RUNTIME}" ]]; then
```
Since we're doing bash, let's use bash test operator.

test/prepare.sh
```diff
@@ -2,18 +2,39 @@
...
+    if podman version > /dev/null 2>&1;  then
```
tuminoid, line 12:
```suggestion
    if command -v podman &> /dev/null;  then
```
More standard way and also using `&>` as its simpler.

test/prepare.sh
```diff
@@ -2,18 +2,39 @@
...
+        CONTAINER_RUNTIME=podman
+    else
+        CONTAINER_RUNTIME=docker
+    fi
+fi
+
+kind_load() {
+    local image="$1"
+    if [ "${CONTAINER_RUNTIME}" = podman ]; then
```
tuminoid, line 21:
```suggestion
    if [[ "${CONTAINER_RUNTIME}" = podman ]]; then
```

test/prepare.sh
elfosardo, line 12: one minor thing, this can be replaced by 'if command -v podman &> /dev/null'

tuminoid (commented):
Repeating the minor nits from #54 as that is probably just building on this.

tuminoid:
Duh, late by seconds.

dtantsur:
I'll address the comments in the 2nd PR, no worries

PR #54, opened 2024-10-28
Clean-up configuration manifests

### LINKED ISSUE #29: Finish built-in deployment scripts
Currently, [the deployment scripts](https://github.com/metal3-io/ironic-standalone-operator/tree/main/config) are an auto-generated mess. In particular, TLS support is commented out, preventing us from using webhooks. This all has to be cleaned up, TLS and webhook support enabled, redundant generated code removed.

dtantsur:
- Use the published image by default
- Enable cert-manager support by default
- Re-enable the webhook port and webhook manifests
- Add a testing overlay to enable all feature gates
- Fix deprecation warnings in kustomizations

Closes: #29

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,72 +1,50 @@
 # Adds namespace to all resources.
 namespace: ironic-standalone-operator-system
-
-# Value of this field is prepended to the
-# names of all resources, e.g. a deployment named
-# "wordpress" becomes "alices-wordpress".
-# Note that it should also match with the prefix (text before '-') of the namespace
-# field above.
 namePrefix: ironic-standalone-operator-
 
-# Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
-
-bases:
+resources:
 - ../crd
 - ../rbac
 - ../manager
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
-# crd/kustomization.yaml
-#- ../webhook
-# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
-#- ../certmanager
+- ../webhook
+- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
 
-patchesStrategicMerge:
+patches:
 # Protect the /metrics endpoint by putting it behind auth.
 # If you want your controller-manager to expose the /metrics
 # endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
-
-
-
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
-# crd/kustomization.yaml
-#- manager_webhook_patch.yaml
-
-# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
-# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
+- path: manager_auth_proxy_patch.yaml
+- path: manager_webhook_patch.yaml
 # 'CERTMANAGER' needs to be enabled to use ca injection
-#- webhookcainjection_patch.yaml
+- path: webhookcainjection_patch.yaml
 
 # the following config is for teaching kustomize how to do var substitution
 vars:
-# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
-#- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
-#  objref:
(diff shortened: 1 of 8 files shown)


tuminoid:
Just noting about GH WOW: do not put issue numbers in titles (we have actions to check against that in other major repos, not here yet it seems), but in summary where it'll trigger the issue auto-closing as you probably wanted.

test/prepare.sh
```diff
@@ -2,18 +2,52 @@
...
 
-IMG="${IMG:-controller:test}"
+IMG="${IMG:-localhost/controller:test}"
 LOGDIR="${LOGDIR:-/tmp/logs}"
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
+CERT_MANAGER_VERSION="${CERT_MANAGER_VERSION:-1.16.1}"
 
 mkdir -p "${LOGDIR}"
 
+if [ -z "${CONTAINER_RUNTIME}" ]; then
```
tuminoid, line 12:
```suggestion
if [[ -z "${CONTAINER_RUNTIME}" ]]; then
```
We're writing bash, so let's use the bash's double brackets.

test/prepare.sh
```diff
@@ -2,18 +2,52 @@
...
+    if podman version > /dev/null 2>&1;  then
```
tuminoid, line 13:
```suggestion
    if command -v podman &> /dev/null;  then
```
Unless you specifically want to run `podman version` for some reason, suggested way avoids executing heavier podman executable, and command is more standard way of checking for tools. Also, `&>` is more efficient way of redirecting everything.

tuminoid (commented):
Minor nits, generally looking good. See previous comment about GH WOW as well.

dtantsur:
>  where it'll trigger the issue auto-closing as you probably wanted

Both places trigger auto-closing, are there other reasons against it?

tuminoid:
> > where it'll trigger the issue auto-closing as you probably wanted
> 
> Both places trigger auto-closing, are there other reasons against it?

You can see from the GH UI that it is not linking the issue, not from this to issue, nor issue to this. Only mentions in summary body works properly.

Also, like mentioned, we even have safeguard actions versus it in CAPM3, IPAM and BMO.

dtantsur:
Ah, hmm, it does work, but only on the commit level (as you can see in https://github.com/metal3-io/ironic-standalone-operator/issues/29). Okay, I'll change.

tuminoid:
> Ah, hmm, it does work, but only on the commit level (as you can see in #29). Okay, I'll change.

In #54 it is also in summary body as `Closes: #29`, I think it is that and not the commit.

Anyways, I'll bring over the PR content verifier workflow here later for this stuff.

dtantsur:
Note: this PR is still going to fail the functional job.

tuminoid (commented):
/lgtm

Rozzii (commented):
/approve

Rozzii (commented):
/lgtm

tuminoid:
> /override functional

You cannot override GH actions. Override works only for Prow based tests.

dtantsur:
Please don't override the test, its failure is real! Previously, the webhook was disabled in tests which allowed us to casually exercise the experimental HA feature. Now that the webhook is enabled, it's blocking that.

We need https://github.com/metal3-io/ironic-standalone-operator/pull/58, then we can enable the HA feature in the tests.

dtantsur:
@tuminoid @Rozzii this is finally ready for review

config/default/kustomization.yaml
```diff
@@ -1,72 +1,50 @@
...
-# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
-# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
+- path: manager_auth_proxy_patch.yaml
+- path: manager_webhook_patch.yaml
 # 'CERTMANAGER' needs to be enabled to use ca injection
-#- webhookcainjection_patch.yaml
+- path: webhookcainjection_patch.yaml
 
 # the following config is for teaching kustomize how to do var substitution
 vars:
```
lentzi90, line 24: Just noting that `vars` are deprecated and succeeded by `replacements`. It is not a 1:1 mapping though so it takes some work to switch. I think we can leave it for now. When the time comes, [this CAPM3 example](https://github.com/metal3-io/cluster-api-provider-metal3/blob/d7d3167b10553c10f8e916aab84d979d3e7303e4/config/default/capm3/kustomization.yaml#L19) should be helpful.

lentzi90 (commented):
/lgtm

config/default/kustomization.yaml
dtantsur: Yeah, I've noticed. I even gave it a try, but it's, as you say, not entirely trivial.

dtantsur: Tracking in https://github.com/metal3-io/ironic-standalone-operator/issues/63

Issue #52, opened 2024-10-22
Bump Golang to 1.23


tuminoid:
Bump Golang to 1.23 before next minor release as 1.22 support will end in 4 months. 

[Release notes](https://tip.golang.org/doc/go1.23) should be checked for any changes that might affect the repo.

NOTE: do not bump `go.mod` versions for the sake of bumping. They shall be bumped only if language features using Go 1.23 are used in code, or by dependencies force it.

/kind feature

PR #55, opened 2024-10-29
bump golang to 1.23.2

iurygregory:
Closes: #52

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Support FROM override
-ARG BUILD_IMAGE=docker.io/golang:1.22.5@sha256:f47a7952d08277f9816459d851319c9041637022f04517388d9f32e384fc2dab
+ARG BUILD_IMAGE=docker.io/golang:1.23.2@sha256:ad5c126b5cf501a8caef751a243bb717ec204ab1aa56dc41dc11be089fafcb4f
 ARG BASE_IMAGE=gcr.io/distroless/static:nonroot@sha256:9ecc53c269509f63c69a266168e4a687c7eb8c0cfd753bd8bfcaa4f58a90876f
 
 # Build the manager binary
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ SHELL = /usr/bin/env bash -o pipefail
 
 .DEFAULT_GOAL:=help
 
-GO_VERSION ?= 1.22.5
+GO_VERSION ?= 1.23.2
 GO := $(shell type -P go)
 # Use GOPROXY environment variable if set
 GOPROXY := $(shell $(GO) env GOPROXY)


iurygregory:
Seems like CI uses 1.22.8 .-.

And I only noticed now the comment in the issue:
NOTE: do not bump go.mod versions for the sake of bumping. They shall be bumped only if language features using Go 1.23 are used in code, or by dependencies force it. 
🤔 So I will think a bit on how to proceed , since this was what I did.

go.mod
```diff
@@ -1,74 +1,75 @@
 module github.com/metal3-io/ironic-standalone-operator
 
-go 1.22.3
+go 1.23.2
```
tuminoid, line 3: As mentioned, this is not wanted. Go.mod go directive shall be the lowest accepted by the dependencies or required by used language features in repo. Our production version is defined in Dockerfile, so bumping this makes no sense and make developer life hard.

go.mod
```diff
@@ -1,74 +1,75 @@
...
 	go4.org/netipx v0.0.0-20231129151722-fdeea329fbba
-	golang.org/x/crypto v0.24.0
-	k8s.io/api v0.30.1
-	k8s.io/apimachinery v0.30.1
-	k8s.io/client-go v0.30.1
-	k8s.io/component-base v0.30.1
-	k8s.io/utils v0.0.0-20230726121419-3b25d923346b
-	sigs.k8s.io/controller-runtime v0.18.3
+	golang.org/x/crypto v0.28.0
+	k8s.io/api v0.31.2
```
tuminoid, line 10: Also keep Go bump minimum to isolate changes. K8s bump for example is not needed in this PR.

tuminoid (changes requested):
This is more of a Go dependency bump, rather than Golang bump that we're after in #52. 

TBH, you could just remove the go directive bumps here, and retitle it, and its 100% valid k8s/controller-runtime bump PR as is, which is also needed. I suggest we go this route and make another PR for golang bump.

go.mod
iurygregory: @tuminoid tks, just to be sure the docker image would be https://hub.docker.com/layers/library/golang/1.23/images/sha256-db933bde327ab6b27f54c29b092ab3bf9276738432ed1bc730b5c9bd98ff33ef?context=explore ?Thanks!

tuminoid: Docker Hub is known to be very confusing about the digest values it shows, so it is easiest to do `docker pull ...` or `podman pull ...` on a amd64 system and use the SHA it gives you. Right now it seems to match what Docker Hub says.

(At time of writing  it is `sha256:ad5c126b5cf501a8caef751a243bb717ec204ab1aa56dc41dc11be089fafcb4f`)

iurygregory: @tuminoid oh ok! Thanks!

tuminoid:
Yup, this fulfills the minimum requirements for a bump.

Did you check the release notes for any changes that could be utilized here, or changes in behavior that needs to be considered?

iurygregory:
> Yup, this fulfills the minimum requirements for a bump.
> 
> Did you check the release notes for any changes that could be utilized here, or changes in behavior that needs to be considered?

I've checked https://tip.golang.org/doc/go1.23 ,  the only thing that caught my eyes was the [Timer changes](https://tip.golang.org/doc/go1.23#timer-changes) , but it shouldn't be a big impact for us.

tuminoid (commented):
If there is language specific changes needed or wanted to be utilized, then
/lgtm

tuminoid (approved):

PR #65, opened 2024-11-06
Add an environment file and a configmap to customize the manager

dtantsur:
Doing so simplifies downstream customizations, e.g. in metal3-dev-env:
https://github.com/metal3-io/metal3-dev-env/pull/1466.

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,2 +1,10 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+
 resources:
 - manager.yaml
+
+configMapGenerator:
+- envs:
+  - manager.env
+  name: ironic-standalone-operator-config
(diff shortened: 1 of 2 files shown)


matthewei:
Hi @dtantsur ,If do that and we have to delete the pod to update the env. Is it really a good way?

Rozzii (commented):
/lgtm

dtantsur:
> If do that and we have to delete the pod to update the env. Is it really a good way?

You're not forced to use the new configmap. This is primarily for consistency with how BMO is configured in its manifests, which in turn will facilitate the effort to add IrSO to metal3-dev-env.

matthewei:
> > If do that and we have to delete the pod to update the env. Is it really a good way?
> 
> You're not forced to use the new configmap. This is primarily for consistency with how BMO is configured in its manifests, which in turn will facilitate the effort to add IrSO to metal3-dev-env.

I see

PR #68, opened 2024-11-08
:seedling: add PR verifier workflow

tuminoid:
As we prepare for MVP and first release of IRSO, let's add PR verifier workflow that enables checking PR title / summary content is compatible with the release machinery scripts.


lentzi90 (commented):
/lgtm

PR #70, opened 2024-11-08
✨ Rework reporting status to the controller

dtantsur:
Currently, the reconciliation code does not distinguish between
potentially transient errors and fatal validation errors (which cannot
be resolved without an update of the resource).

This PR adds a Status object that is loosely modelled after the
provisioner's Result in BMO. More helpers are added to avoid repetition.

As a bonus, we now distinguish between two states: resources are being
updated and we are waiting for resources to get ready.


Rozzii (commented):
/lgtm

PR #73, opened 2024-11-18
🐛 Fix invalid names for webhooks

dtantsur:
I have no idea what kb.io is, or why the webhooks got generated this
way. Fix to provide reasonable names.


tuminoid (commented):
I have been wondering about those, always forgot to ask :) 

/approve

PR #75, opened 2024-11-19
✨ Add support for overrides for generated pods

dtantsur:
This change adds a new field Overrides to both Ironic and
IronicDatabase. This field allows advanced operators and downstream
consumers to inject containers, annotations, labels and environment
variables into the generated deployments/daemon sets.

A new feature gate Overrides is added to protect this feature.


pkg/ironic/utils_test.go
tuminoid, line 1: Would we need to have a test case for checking mixing Env and EnvFrom overrides?

tuminoid (commented):
Couple questions, generally looking solid.

/approve

dtantsur:
True. I won't have time to work on it in the near future, so closing.

PR #78, opened 2024-11-21
:seedling: Change webhook secretName to avoid conflicts

elfosardo:
We change the secretName of the certificate assigned to the
webhhok from a generic webhook-server-cert to include a prefix
"irso" to help avoiding confusion with other certs.

### DIFF
--- a/config/certmanager/certificate.yaml
+++ b/config/certmanager/certificate.yaml
@@ -36,4 +36,4 @@ spec:
   issuerRef:
     kind: Issuer
     name: selfsigned-issuer
-  secretName: webhook-server-cert # this secret will not be prefixed, since it's not managed by kustomize
+  secretName: irso-webhook-server-cert # this is used as-is as the secret is not managed by kustomize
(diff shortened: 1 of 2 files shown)


config/certmanager/certificate.yaml
```diff
@@ -36,4 +36,4 @@ spec:
   issuerRef:
     kind: Issuer
     name: selfsigned-issuer
-  secretName: webhook-server-cert # this secret will not be prefixed, since it's not managed by kustomize
+  secretName: irso-webhook-server-cert # this secret will not be prefixed, since it's not managed by kustomize
```
tuminoid, line 39: Is the comment still valid?

dtantsur:
Please rebase to pick up the CI fix

config/certmanager/certificate.yaml
tuminoid: Its good to have it then, maybe reworded `used as-is, secret is not managed by Kustomize` or something like that

tuminoid:
Can you also provide PR description?

tuminoid (commented):
/lgtm

PR #80, opened 2024-11-22
🐛 Fix image override in the Makefile, collect more logs

dtantsur:
Currently, it overrides a non-existing image location.

Update log collection with information I used to debug this:

Make sure the Ironic resource is collected even if it never finishes
provisioning. Also collect Deployment/DaemonSet. Collect the manager's
deployment and pods in the end of a run.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #82, opened 2024-11-24
⚠️ Rename the API group from metal3.io to ironic

dtantsur:
Initially, I used metal3.io like BMO does, but it does not look correct
and causes absurd results when operator-sdk is used. We were not
supposed to duplicate the domain in the group.


dtantsur:
/hold

I'm still not 100% sure

tuminoid:
/cc @kashifest 
PTAL and express opinion as well.

dtantsur:
/hold cancel

See also https://groups.google.com/u/1/g/metal3-dev/c/8ez2LCM9cng

tuminoid (commented):
/approve

PR #83, opened 2024-11-25
🐛 Handle surrounding newlines and spaces in user names and passwords

dtantsur:
It's a common error to include a newline in secrets with user names and
passwords. So common, that I made it in my metal3-dev-env patch.

Log both values on user name mismatch.

### DIFF
--- a/pkg/ironic/secrets.go
+++ b/pkg/ironic/secrets.go
@@ -45,10 +45,14 @@ func generatePassword() ([]byte, error) {
 	return password, nil
 }
 
-func generateHtpasswd(user, password string) (string, error) {
+func normalizeSecretValue(value []byte) string {
+	return strings.Trim(string(value), " \n\r\t")
+}
+
+func generateHtpasswd(userBytes, passwordBytes []byte) (string, error) {
 	// A common source of errors: an accidental line break after a password
-	user = strings.Trim(user, " \n\r")
-	password = strings.Trim(password, " \n\r")
+	user := normalizeSecretValue(userBytes)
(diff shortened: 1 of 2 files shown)


pkg/ironic/secrets.go
```diff
@@ -45,10 +45,14 @@ func generatePassword() ([]byte, error) {
 	return password, nil
 }
 
-func generateHtpasswd(user, password string) (string, error) {
+func normalizeSecretValue(value []byte) string {
+	return strings.Trim(string(value), " \n\r")
```
tuminoid, line 49: While at it, maybe add `\t` to be trimmed out as well?

dtantsur: A good idea

pkg/ironic/secrets_test.go
```diff
@@ -52,12 +52,32 @@ func TestIsValidUser(t *testing.T) {
...
+			Password: "pa$$w0rd\n",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Scenario, func(t *testing.T) {
+			result, err := generateHtpasswd([]byte(tc.User), []byte(tc.Password))
+			assert.NoError(t, err)
+			user, password, ok := strings.Cut(result, ":")
+			assert.Truef(t, ok, "%s does is not separated with a colon", result)
```
tuminoid, line 76:
```suggestion
			assert.Truef(t, ok, "%s is not separated with a colon", result)
```
Missed this earlier, funny typo.

tuminoid (commented):
/approve

Issue #91, opened 2024-11-27
Settle down on the naming condition: somethingRef vs somethingName


dtantsur:
I've read somewhere (forgot where) that using SomethingRef of type corev1.LocalObjectReference is the preferred way to link to other objects.

On one hand, it allows adding Kind/Namespace later if needed without a breaking change.

On the other, it feels ugly, and does not seem to be used consistently in other major operators. [Postgres uses SecretName/CaSecretName](https://postgres-operator.readthedocs.io/en/latest/user/#custom-tls-certificates) for TLS. [CAPI operator has AdditionalManifestsRef](https://github.com/kubernetes-sigs/cluster-api-operator/blob/e70a9ee10dee176826dc29867c099e9026607e15/api/v1alpha2/provider_types.go#L67). BMO is inconsistent: ConsumerRef, NetworkData, PreprovisioningNetworkDataName.

We need to decide before the MVP.

/triage accepted

dtantsur:
/cc @tuminoid @lentzi90 @elfosardo @Rozzii 

Looking for feedback here.

tuminoid:
Ref suffix is more flexible for future changes, or when it refers across namespaces, but when relationship is within same namespace and of known type, Name seems to be used.

I quickly grepped the repos I have cloned (opinionated sample of course) for hits on Name: and Ref: in manifests and its roughly split 50/50, even within projects. Almost all repos have both.

Lets also ping @adilGhaffarDev and @kashifest for this.

dtantsur:
Similarly, I'm starting to think that we need simple names for links to IronicDatabase, TLS and credentials secrets, and so on.

kashifest:
Not an expert in this regard but I am leaning towards Refs (although I don't have any strong points against names.)

dtantsur:
Hmm, we got a split here. I've just posted a WIP patch using names, so that you can compare two approaches.

adilGhaffarDev:
To my understanding, `ObjectReference` is used when it is not clearwhat will be in the field and depends on the outside entity. For example in CAPI `ObjectReference` is mostly used when something that is being set to a field is coming from outside. like here:
https://github.com/kubernetes-sigs/cluster-api/blob/2719ae3cc94bbd72dbcbfecc254615487693de86/api/v1beta1/machinehealthcheck_types.go#L110 
RemediationTemplate is provided by the infrastructure provider.
[...]

PR #92, opened 2024-11-28
⚠️ Rework naming of references in the API

dtantsur:
As discussed in
https://github.com/metal3-io/ironic-standalone-operator/issues/91, lets
use simpler reference names and reserve the SomethingRef format for when
the field can refer to different kinds or at least namespaces.

This is quite a breaking change but I don't see a lot of value in going
through deprecation since it's going to significantly complicate the
code nobody should be using in production yet.

Closes: #91 


dtantsur:
/hold

I'll create a corresponding metal3-dev-env change.

dtantsur:
https://github.com/metal3-io/metal3-dev-env/pull/1478 pins IrSO to avoid breaking metal3-dev-env with this.

tuminoid:
> This is quite a breaking change but I don't see a lot of value in going through deprecation since it's going to significantly complicate the code nobody should be using in production yet.

We absolutely should not make life difficult for ourselves here, this is pre-MVP.

tuminoid (commented):
Looks much cleaner to me.

/approve

elfosardo:
best do a change like this now

/lgtm

Issue #71, opened 2024-11-11
Add golangci-lint


tuminoid:
Adding golangci-lint with as many linters enabled as possible is needed before releasing to ensure code quality.

As IRSO is using Golang 1.23, the required golangci-lint version is 1.60.1 or newer. Golangci-lint v1.60.x brings in new linters, so if/when adapting golangci-lint from BMO/CAPM3, the linter's allow list need to be checked to be up to date.

elfosardo:
Just had some time and put together a quick PR for this

PR #94, opened 2024-12-03
:seedling: Add golangci-lint

elfosardo:
Fix some errors but disable more intrusive linters for the time
being.

Closes https://github.com/metal3-io/ironic-standalone-operator/issues/71

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,33 @@
+name: golangci-lint
+
+on:
+  pull_request:
+    types: [ opened, edited, synchronize, reopened ]
+
+jobs:
+  golangci-lint:
+    name: lint
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        working-directory:
+        - ""
+        - api
+        - internal/controller
+        - test
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Calculate go version
+      id: vars
+      run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
+    - name: Set up Go
+      uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # v5.1.0
+      with:
+        go-version: ${{ steps.vars.outputs.go_version }}
+    - name: golangci-lint-${{matrix.working-directory}}
+      uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6.1.1
+      with:
+        version: v1.60.3
+        working-directory: ${{matrix.working-directory}}
+        args: --timeout=10m
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,122 @@
+run:
+  deadline: 10m
+linters:
+  disable-all: true
+  enable:
+  - asasalint
+  - asciicheck
+  - bidichk
+  - bodyclose
+  #- containedctx
+  - copyloopvar
+  - dogsled
+  - dupword
+  - durationcheck
+  #- errcheck
+  - errchkjson
+  #- gci
+  - ginkgolinter
+  - goconst
+  #- gocritic
+  #- godot
+  - gofmt
+  - goimports
+  - goprintffuncname
+  #- gosec
+  #- gosimple
+  - govet
+  #- importas
+  - ineffassign
+  - loggercheck
+  - misspell
+  #- nakedret
+  #- nilerr
+  - noctx
+  - nolintlint
+  - nosprintfhostport
+  - prealloc
+  #- predeclared
+  #- revive
+  - rowserrcheck
+  - staticcheck
+  #- stylecheck
+  - thelper
+  - typecheck
+  - unconvert
+  #- unparam
+  #- unused
+  - usestdlibvars
+  #- whitespace
+  # Run with --fast=false for more extensive checks
+  fast: true
+linters-settings:
+  gosec:
+    go: "1.23"
+    severity: medium
+    confidence: medium
+    concurrency: 8
+  importas:
+    no-unaliased: true
+    alias:
+    # Kubernetes
+    - pkg: k8s.io/api/core/v1
+      alias: corev1
+    - pkg: k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1
+      alias: apiextensionsv1
+    - pkg: k8s.io/apimachinery/pkg/apis/meta/v1
+      alias: metav1
+    - pkg: k8s.io/apimachinery/pkg/api/errors
+      alias: k8serrors
+    - pkg: k8s.io/apimachinery/pkg/util/errors
+      alias: kerrors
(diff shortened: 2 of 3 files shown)


elfosardo:
@tuminoid @dtantsur  I took the config from BMO which has the same version now, so the linters are the same, I see quite some errors, do we want to modify the linters or fix the errors? :smile:

tuminoid:
> @tuminoid @dtantsur I took the config from BMO which has the same version now, so the linters are the same, I see quite some errors, do we want to modify the linters or fix the errors? 😄

When we introduced golangci-lint in BMO/CAPM3 etc, we first enabled only the passing set of linters, commented the others out. Then PR by PR enabled those linters with required code fixes. This way the patches are reasonable and its clear why the changes are made.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,39 @@
...
+      fail-fast: false
+      matrix:
+        working-directory:
+        - ""
+        - api
+        - controllers
+        - test
+    steps:
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - name: Install libvirt
```
tuminoid, line 23: This is not needed in IRSO

.golangci.yaml
```diff
@@ -0,0 +1,183 @@
+run:
+  deadline: 10m
+  build-tags:
```
tuminoid, line 3: These are BMO tags

.golangci.yaml
```diff
@@ -0,0 +1,183 @@
...
+  - e2e
+  - vbmctl
+linters:
+  disable-all: true
+  enable:
```
tuminoid, line 8: Not too bad list of failures, we'd need to comment out:
- errcheck
- gci
- godot
- gosec
- revive
- unconvert
- unused
- whitespace
- nakedret

and out of those, we could turn back on:
- whitespace
- unused
- unconvert
- gosec

with a single line fix for each in this PR even, leaving only 5 disabled.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,39 @@
+name: golangci-lint
+
+on:
+  push:
```
tuminoid, line 4: Trigger should be

```yaml
on:
  pull_request:
    types: [ opened, edited, synchronize, reopened ]
```

as seen in CAPM3. BMO linter is part of more strucutred actions setup, so it is not good example for triggers.

elfosardo:
\o/

pkg/ironic/ironic.go
```diff
@@ -126,28 +126,30 @@ func ensureIronicService(cctx ControllerContext, ironic *metal3api.Ironic) (Stat
...
 		return
 	}
 
-	if validationErr := metal3api.ValidateIronic(&ironic.Spec, nil); validationErr != nil {
+	validationErr := metal3iov1alpha1.ValidateIronic(&ironic.Spec, nil)
+	if validationErr != nil {
 		status = Status{Fatal: validationErr}
-		return
+		//lint:ignore nilerr reason
+		return // ignore
```
tuminoid, line 152:
```suggestion
		return //nolint:nilerr
```
That should ignore it in more concise and consistent way.

Though this looks like it would also warrant proper fix. Returning nothing instead of two values is sketchy. It can be fixed in later PRs as well.

tuminoid (commented):
Thanks for adding it, looking great! One minor nit.

/approve

pkg/ironic/ironic.go
elfosardo: yes, honestly I forgot to change it with the one liner!
And I agree this should be fixed and not ignored, but it's not doing too much harm AFAICS
I'm planning to take care of the rest of the stuff we ignored commenting out the linters, I'll add this one to the list

.golangci.yaml
```diff
@@ -0,0 +1,180 @@
...
+    - pkg: k8s.io/apimachinery/pkg/util/errors
+      alias: kerrors
+    - pkg: k8s.io/component-base/logs/api/v1
+      alias: logsv1
+    # Controller Runtime
+    - pkg: sigs.k8s.io/controller-runtime
+      alias: ctrl
+    # IrSO
+    - pkg: github.com/metal3-io/ironic-standalone-operator/api/v1alpha1
+      alias: metal3iov1alpha1
```
dtantsur, line 79: Btw this alias is no longer correct. After I've fixed the API group to "ironic", Kubebuilder generates `ironicv1alpha1`. We can merge this as it is and clean up later.

.golangci.yaml
```diff
@@ -0,0 +1,180 @@
...
+  unused:
+    go: "1.23"
+issues:
+  skip-dirs:
+  - mock*
+  skip-files:
+  - "zz_generated.*\\.go$"
+  - ".*conversion.*\\.go$"
+  exclude-rules:
+  - path: test/e2e
```
dtantsur, line 110: We don't have an e2e directory here.

internal/controller/ironic_controller.go
```diff
@@ -35,7 +35,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/go-logr/logr"
-	metal3api "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
+	metal3iov1alpha1 "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
```
dtantsur, line 38: Please don't make this change, I literally reverted this in BMO already.

internal/controller/ironicdatabase_controller.go
```diff
@@ -30,7 +30,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/go-logr/logr"
-	metal3api "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
+	metal3iov1alpha1 "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
```
dtantsur, line 33: Same here and in other files

test/suite_test.go
```diff
@@ -240,24 +240,27 @@ func CollectLogs(namespace string) {
 
 		writeYAML(&pod, namespace, pod.Name, "pod")
 
+		var podLogs io.ReadCloser
```
dtantsur, line 243: This entire change does not look valid to me. Especially using `defer` is the correct and idiomatic way to handle this case.

dtantsur (changes requested):
Thanks for taking on this work!

I'd prefer we only enable linters that only pass and enable the rest separately. Many of the changes made here don't look valid to me.

.golangci.yaml
elfosardo: I've changed it to ironicv1alpha1
we'll clean up when enabling the importas linter

test/suite_test.go
elfosardo: considering that this is a test file it's probably not a big deal and we may ignore it, but my understanding is that using defer inside for loop is not recommended in general, it's exposed by the gocritic linter https://go-critic.com/overview.html#deferinloop

elfosardo:
@dtantsur I've simplified a lot this change, we can work on single linters one by one

dtantsur (approved):
/lgtm

Thanks!

PR #110, opened 2025-01-13
🌱 Label pods/deployments with a version and clean up selector labels

dtantsur:
Add a label with the Ironic version to all Ironic deployments and pods,
so that we can apply version-specific logic down the road.

Clean up the current labels: use the same prefix as the API group and do
not reuse the same label for Ironic and database to avoid conflicts.


tuminoid (commented):
/approve

Good improvement.

Rozzii (commented):
/lgtm

PR #112, opened 2025-01-14
🌱 CI: implement a test for TLS on Ironic API

dtantsur:
Also make sure that environment-specific parameters come from prepare.sh
as variables rather than being hardcoded.

### DIFF
--- a/test/prepare.sh
+++ b/test/prepare.sh
@@ -7,6 +7,8 @@ LOGDIR="${LOGDIR:-/tmp/logs}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
 CERT_MANAGER_VERSION="${CERT_MANAGER_VERSION:-1.16.1}"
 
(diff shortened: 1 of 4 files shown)


test/prepare.sh
```diff
@@ -7,6 +7,8 @@ LOGDIR="${LOGDIR:-/tmp/logs}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-}"
 CERT_MANAGER_VERSION="${CERT_MANAGER_VERSION:-1.16.1}"
 
+. "$(dirname $0)/testing.env"
```
tuminoid, line 10:
```suggestion
. "$(dirname "$0")/testing.env"
```

tuminoid (commented):
/approve

/hold
Just unhold if you don't want to fix the comment now, but later.

dtantsur:
/unhold

I'll follow-up shortly, would prefer to get the rest of the patch in.

Issue #89, opened 2024-11-27
Replace README with a proper user guide


dtantsur:
By MVP, we need some documentation on how users install and use the operator.
README should only contain the very basic information.

/kind documentation
/triage accepted

dtantsur:
User guide: https://github.com/metal3-io/metal3-docs/pull/493

PR #113, opened 2025-01-16
🌱 Update README, drop ARCHITECTURE

dtantsur:
All the actually useful information has been migrated to the user guide.
What is left mostly refers to experimental bits which should be
documented properly once they are ready.

Closes: #89


tuminoid (commented):
/approve

PR #122, opened 2025-02-05
🌱 Enable functional test to be used on ironic-image/mariadb-image

dtantsur:
This change adds new environment variables to control which Ironic and
MariaDB images are used in the tests. The goal is to run the same suite
on both repositories to avoid regressions.


lentzi90 (commented):
/lgtm

PR #129, opened 2025-02-11
🌱 Functional tests: improve versions test, add upgrades

dtantsur:
Currently, the relevant test just sets the Version field without any
further validation. Here I'm relying on API versions to check that
the correct version is installed since latest and 27.0 have diverged.

Try upgrading from 27.0 to latest.

Clean up redundant constants in version.go.

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -34,6 +34,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/openstack/baremetal/apiversions"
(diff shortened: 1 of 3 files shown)


test/suite_test.go
```diff
@@ -54,6 +55,12 @@ import (
 	metal3api "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
 )
 
+// NOTE(dtantsur): latest is now at least 1.95
+const apiVersionIn270 = "1.94"
```
tuminoid, line 59: Can we document in a comment what exactly these versions are representing and where are they coming from? In Metal3 context, they're completely alien.

tuminoid (commented):
/lgtm

Looks good to me, Ironic version numbering comment is spot on, thanks.

PR #135, opened 2025-02-12
🐛 Bump Golang to 1.23.6 to pick up the fix for GO-2025-3447

dtantsur:


tuminoid (commented):
/approve

tuminoid:
> Summary: Timing sidechannel for P-256 on ppc64le in crypto/internal/nistec

I don't think we support ppc64le arch, but Go bumping never hurts.

tuminoid:
> > I don't think we support ppc64le arch, but Go bumping never hurts.
> 
> Yeah, I don't want various scanners to scream at us :)

Yeah, the code path is in use, so it is worth bumping to silence them.

lentzi90 (commented):
/lgtm

Issue #125, opened 2025-02-06
Prevent further reconciliation if the Version is not supported


dtantsur:
Right now, the webhook prevents a user from setting a wrong version. However, as the operator evolves, we'll eventually start removing Ironic versions from it. This may leave users with Ironic resources that cannot be reasonably reconciled (if I read the code correctly, we may do something absurdly wrong in this case). We need to detect this situation very early in the reconciliation cycle and set an error condition on the Ironic object. The message should recommend to upgrade Ironic using an older version of the operator.

/kind bug
/triage accepted
/help

PR #138, opened 2025-02-17
:warning: Harden the validation around the ironic version

dtantsur:
Currently, the only validation exists on the webhook level. When the
webhook is disabled or the resource is upgraded from an older version,
the reconciliation may start with a wrong version and may fail or even
panic later on.

To avoid it, introduce validation on both CLI and controller levels.
Add a new Version structure that can be used later for version-specific
logic.

Note that this change has an upgrade impact: an Ironic with a no longer
supported version now must be updated before an upgrade.

Closes: #125


lentzi90 (commented):
/lgtm

PR #140, opened 2025-02-23
:warning: Finish transition to typed Version

dtantsur:
Currently, an incomplete VersionInfo is passed around until it's fully validated when the Ironic object is available.
Furthermore, Version is passed as a string, requiring conversion for any comparisons (there will be many of them in the near future once we start writing version-specific logic).

This series of commits refactors the handling of Version and VersionInfo so that
1. An invalid VersionInfo object is never constructed
2. A potentially invalid version string is never passed around
3. Defaults for VersionInfo are applied when constructing it, not when overriding the defaults from the CR
4. `init`-time validation of versions is no longer required

A comparison function is added to Version for the future use.

These changes are technically breaking since the types of public constants and types are changed.

### DIFF
--- a/api/v1alpha1/version.go
+++ b/api/v1alpha1/version.go
@@ -1,37 +1,63 @@
 package v1alpha1
 
 import (
+	"cmp"
 	"fmt"
 	"slices"
 	"strconv"
 	"strings"
 )
 
+const versionLatestString = "latest"
+
 type Version struct {
 	Major, Minor int
+}
+
+func (v Version) Compare(other Version) int {
+	if v.IsLatest() {
+		if other.IsLatest() {
+			return 0
+		}
+		return 1
+	}
+
+	if v.Major != other.Major {
+		return cmp.Compare(v.Major, other.Major)
+	}
+
+	return cmp.Compare(v.Minor, other.Minor)
 }
 
 func (v Version) IsLatest() bool {
 	return v.Major == 0
 }
 
+func (v Version) String() string {
+	if v.IsLatest() {
+		return versionLatestString
+	}
+
+	return fmt.Sprintf("%d.%d", v.Major, v.Minor)
+}
+
 func ParseVersion(version string) (Version, error) {
-	if version == VersionLatest {
+	if version == versionLatestString {
 		return Version{}, nil
 	}
 
 	versionSplit := strings.SplitN(version, ".", 2)
 	if len(versionSplit) != 2 {
-		return Version{}, fmt.Errorf("invalid version %s, expected X.Y", version)
+		return Version{}, fmt.Errorf("invalid version %s, expected MAJOR.MINOR", version)
 	}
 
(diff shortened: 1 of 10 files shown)


api/v1alpha1/version.go
```diff
@@ -1,22 +1,48 @@
...
+			return 0
+		}
+		return 1
+	}
+
+	if majorCmp := cmp.Compare(v.Major, other.Major); majorCmp != 0 {
+		return majorCmp
+	}
+
+	return cmp.Compare(v.Minor, other.Minor)
```
tuminoid, line 29:
```suggestion
    if v.Major != other.Major {
        return cmp.Compare(v.Major, other.Major)
    }
    return cmp.Compare(v.Minor, other.Minor)
```
Simpler code IMO.

cmd/main.go
```diff
@@ -147,11 +149,10 @@ func main() {
...
 	setupLog.Info("enabling features", "FeatureGate", metal3iov1alpha1.CurrentFeatureGate.String())
 
-	if versionInfo.InstalledVersion != "" {
-		if err := metal3iov1alpha1.ValidateVersion(versionInfo.InstalledVersion); err != nil {
-			setupLog.Error(err, "invalid ironic-version")
-			os.Exit(1)
-		}
+	versionInfo, err := ironic.NewVersionInfo(ironicImages, ironicVersion, databaseImage)
+	if err != nil {
+		setupLog.Error(err, "invalid ironic-version")
```
tuminoid, line 154:
```suggestion
		setupLog.Error(err, "invalid ironic-version: %q", ironicVersion)
```
Personally I prefer invalid values shown in error messages.

internal/controller/ironic_controller.go
```diff
@@ -122,7 +122,7 @@ func (r *IronicReconciler) handleIronic(cctx ironic.ControllerContext, ironicCon
 	}
 	cctx.VersionInfo = versionInfo
 
-	actuallyRequestedVersion := cctx.VersionInfo.InstalledVersion
+	actuallyRequestedVersion := cctx.VersionInfo.InstalledVersion.String()
```
tuminoid, line 125: Could we assign `cctx.VersionInfo.InstalledVersion.String()` to var, as it is converted twice (L125 and L155) and is hard to read too?

dtantsur: You mean, reuse actuallyRequestedVersion on line 155? Can do it.

cmd/main.go
dtantsur: You'll get it through `err`, the initial error message already includes the invalid value

internal/controller/ironic_controller.go
tuminoid: Like said, it was hard to read :laughing:

tuminoid (commented):
/lgtm

PR #141, opened 2025-02-24
Update supported versions for ironic-image release-28.0

elfosardo:


test/suite_test.go
```diff
@@ -656,14 +657,48 @@ var _ = Describe("Ironic object tests", func() {
...
 
+		By("upgrading to Ironic 28.0")
+
+		patch := client.MergeFrom(ironic.DeepCopy())
+		ironic.Spec.Version = "28.0"
+		err = k8sClient.Patch(ctx, ironic, patch)
+		Expect(err).NotTo(HaveOccurred())
+
+		ironic = WaitForUpgrade(name, "27.0", "28.0")
+		VerifyIronic(ironic, TestAssumptions{})
```
dtantsur, line 668: You could use `TestAssumptions{maxAPIVersion: apiVersionIn280}` here for an extra check

elfosardo: makes sense
I'll add it

Rozzii (commented):
/approve

PR #148, opened 2025-03-06
:seedling: add go ecosystem to dependabot config

tuminoid:
Add go ecosystem to dependabot config, so we get automated updates for go modules.


Rozzii (commented):
/approve

PR #150, opened 2025-03-06
🌱 Check the httpd server in the functional tests

dtantsur:
Make sure it's running, using TLS if enabled and that the IPA ramdisk
gets downloaded there (or not, when the downloader is disabled).


tuminoid (commented):
/lgtm

Rozzii (commented):
/approve

PR #151, opened 2025-03-06
🐛 Fix probes for httpd when the downloader is disabled

dtantsur:
The current probes rely on /images path which is created by the
downloader, so the probe fails when it's disabled.

Even worse, we're relying on an HTTP redirect from /images to /images/ being
interpreted as success by cURL. Adding -L uncovers the issue: /images/
returns HTTP 404. So, use -L to avoid false positives and switch to
a known path that the downloader creates.

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -27,6 +27,8 @@ const (
 	authDir   = "/auth"
 	certsDir  = "/certs"
 	sharedDir = "/shared"
+
+	knownExistingPath = "/images/ironic-python-agent.kernel"
 )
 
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
/approve

Rather classic curl usage story :)

Should we maybe make an improvement issue of not execing binaries, but using net/http, since this is not test code?

dtantsur:
> Should we maybe make an improvement issue of not execing binaries, but using net/http, since this is not test code?

You mean, the URL probe instead of the Exec one? I don't think that allows TLS settings, this is why I'm using curl for now.

PR #161, opened 2025-03-11
:seedling: add dependabot workflow

tuminoid:
In addition to dependabot config that allows dependabot to create PRs for bumps, for Go ecosystem we also need a dependabot workflow that after bumps also adds commit with result of "make modules", and "make generate" to each dependabot commit.

I forgot to add this when adding dependabot config for the Go.

### DIFF
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -1,0 +1,51 @@
+name: dependabot
+
+on:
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
/lgtm

Rozzii (commented):
/approve

Issue #142, opened 2025-02-26
Reconsider having explicit support for deploying MariaDB


dtantsur:
There is presumably no shortage of ways to deploy MariaDB that are better supported than mariadb-image. [The official documentation](https://mariadb.com/kb/en/kubernetes-operators-for-mariadb/) refers to [mariadb-operator](https://github.com/mariadb-operator/mariadb-operator). By relying on an external operator we can get HA and upgrades for free.

The proposal here is to deprecate and remove `IronicDatabase` CR, mariadb-image repository and start accepting a link to an externally managed MariaDB, e.g. like this:

```yaml
spec:
  database:
    serviceName: mydb-primary
    name: myironic
    credentialsName: secret-with-credentials
    tlsName: secret-with-tls-cert-or-ca
```

Open questions:
- Any reasons to keep mariadb-image?
- Are we okay with referring to a service?
- Are we okay with not supporting databases from other namespaces?
- Do we need to support IP:PORT for databases that are not deployed in Kubernetes?
- Anything else we need there?

/kind feature
/triage accepted

dtantsur:
cc @tuminoid

tuminoid:
> Open questions:
> 
> * Any reasons to keep mariadb-image?

We need to make a design choice here. IMO just doing repackaging of mariadb in a container is just dead weight to us, and we really don't want to be responsible for a DB container in production. We do need to offer something for quick testing and prototyping, and IMO Sqlite in Ironic is good for that. Any serious production environment should use properly managed DB instance.

I would be happy to see mariadb-image retired.

> * Are we okay with referring to a service?

If you're asking if were ok referring to mariadb-operator, that is the Mariadb corps official operator. We should have examples how to achieve the mariadb-image parity with it (single container with equal security posture), ie. some sort of migration guide.

> * Are we okay with not supporting databases from other namespaces?
> * Do we need to support IP:PORT for databases that are not deployed in Kubernetes?

IMO, we should not care where the database is, as long as we can connect to it. Forcing user to install DB in our namespace for just this is not good practice. We should accept whatever address, creds user configures as long as it is supported version of Mariadb. If they happen to run a huge cluster outside k8s, why do we care? If user wants to point to AWS RDS, go for it. As long as it fulfills the requirements that is. Ironic networking is probably the most limiting thing here?

> * Anything else we need there?

[...]

PR #163, opened 2025-03-11
⚠️ Deprecate the IronicDatabase API and spec.databaseName on Ironic

dtantsur:
We should not be in the business of managing databases. Our
mariadb-image has known issues in the initialization code, does not
support HA and cannot do upgrades. There are 3rd party solutions that
are much more production-ready, just use them.

Part of: #142

### DIFF
--- a/api/v1alpha1/ironic_types.go
+++ b/api/v1alpha1/ironic_types.go
@@ -238,6 +238,8 @@ type IronicSpec struct {
 
 	// DatabaseName is a reference to the IronicDatabase object.
 	// If missing, a local SQLite database will be used. Must be provided for a highly available architecture.
+	//
+	// Deprecated: the IronicDatabase API is deprecated and will be removed soon in favour of explicit connection parameters.
 	// +optional
 	DatabaseName string `json:"databaseName,omitempty"`
 
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -123,3 +123,9 @@ issues:
   - EXC0002 # include "missing comments" issues from golangci-lint
   max-issues-per-linter: 0
   max-same-issues: 0
+  exclude-rules:
+  # NOTE(dtantsur): IronicDatabase is deprecated but our own code still supports it.
+  # Remove this exclusion when it's entirely deleted.
+  - linters:
+      - staticcheck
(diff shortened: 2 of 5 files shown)


tuminoid (commented):
/approve

api/v1alpha1/ironic_types.go
```diff
@@ -238,6 +238,7 @@ type IronicSpec struct {
 
 	// DatabaseName is a reference to the IronicDatabase object.
 	// If missing, a local SQLite database will be used. Must be provided for a highly available architecture.
+	// Deprecated: the IronicDatabase API is deprecated and will be removed soon in favour of explicit connection parameters.
```
lentzi90, line 241: I have seen in [CAPI PRs](https://github.com/kubernetes-sigs/cluster-api/pull/10798#discussion_r1652945411) that they always put an empty line before the deprecation. It is supposedly needed for some tooling (linters?) to pick up on it.
Not sure how important it is for us

elfosardo: yep, according to godoc it should be a new paragraph, so one blank line before the statement is needed to separate it

dtantsur: TIL!

tuminoid: Btw, not whole blank line, but commented empty line. I just fixed issue like that in CAPM3.
https://github.com/metal3-io/cluster-api-provider-metal3/pull/2385

dtantsur: Mmm, yeah, it works too well: now the linter fails with a dozen of deprecation warnings all over the place :( Can we merge it as it is? I don't intend this to be a Go-level deprecation, more of API-level.

elfosardo: I think that's reasonable
if needed we can maybe fix that and other deprecation statements in a different patch

lentzi90: I would suggest doing the deprecation properly and then adding it as ignored in the linter config, like so:
https://github.com/kubernetes-sigs/cluster-api/blob/56879a0640862d6652df455503acfd5b36fd1565/.golangci.yml#L242-L255
Otherwise it will be confusing and easy to forget later

dtantsur:
@lentzi90 good pointer, done!

PR #168, opened 2025-03-13
WIP online data migrations via a Job

dtantsur:


dtantsur:
This change finally works and is passing the CI 🎉

api/v1alpha1/version.go
```diff
@@ -41,6 +41,10 @@ func (v Version) String() string {
 	return fmt.Sprintf("%d.%d", v.Major, v.Minor)
 }
 
+func (v Version) HasUpgradeScripts() bool {
+	return v.Compare(Version280) > 0
```
tuminoid, line 45: Why > Version280 rather than >= Version290?

config/rbac/role.yaml
```diff
@@ -48,6 +48,18 @@ rules:
   - patch
   - update
   - watch
+- apiGroups:
+  - batch
+  resources:
+  - jobs
+  verbs:
```
tuminoid, line 55: Do we really use all the verbs? We should list only verbs we actually use.

dtantsur: Probably don't need list, but the whole CRUD is used.

pkg/ironic/upgrades.go
```diff
@@ -0,0 +1,143 @@
...
+
+	// TODO(dtantsur): remove this when ironic-image <= 28.0 is not supported
+	if !toVersion.HasUpgradeScripts() {
+		cctx.Logger.Info("not running upgrade scripts: the new version does not support them", "From", fromVersion, "To", toVersion.String())
+		return ready()
+	}
+
+	job := &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-%s-%s-to-%s", ironic.Name, phase, fromVersion, cctx.VersionInfo.InstalledVersion),
```
lentzi90, line 109: Is there any risk that the same job might be needed again later, e.g. if doing a rollback and then again upgrade? If there is such a risk, we could consider using `generateName` instead and/or [adding a TTL](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) (that might be nice anyway).

lentzi90 (commented):
Looks nice!
One question/note below

pkg/ironic/upgrades.go
dtantsur: Ironic does not support downgrades, but adding a TTL is probably a good idea.

api/v1alpha1/version.go
tuminoid: It does exist now, #213 was merged.

tuminoid (commented):
/approve

Issue #67, opened 2024-11-07
Finish unit tests for the validating webhook


dtantsur:
https://github.com/metal3-io/ironic-standalone-operator/pull/66 adds some tests, more are needed (cover DHCP, database logic, etc).

/triage accepted
/help

dtantsur:
@matthewei hi! We're determining, which changes will go into the MVP release. Do you have any updates here?

PR #175, opened 2025-03-21
🌱 Migrate to new style webhooks from controller-runtime 0.17

dtantsur:
What we're doing no longer works in controller-runtime 0.20.
This change rearranges webhooks to follow the example of Kubebuilder [1]

- The webhook code is migrated to internal/webhook/v1alpha1
- The validation code is migrated to pkg/ironic since it's also used
  there for reconciliation-time validation
- The IronicDatabase webhook does nothing and is removed here

The benefit of the new arrangement is that we can create an api Go
module without a dependency on controller-runtime (as well as the
packages we use for validation).

[1] https://github.com/kubernetes-sigs/kubebuilder/pull/4150/files


adilGhaffarDev:
LGTM, good change I will push similar changes in CAPM3 and IPAM too. 
It would be nice to add tests too for the webhooks. Perhaps something that can be done in the future.

dtantsur:
> It would be nice to add tests too for the webhooks. Perhaps something that can be done in the future.

I agree, there is even a tracker for that: https://github.com/metal3-io/ironic-standalone-operator/issues/67

tuminoid (commented):
/approve

tuminoid:
/lgtm

I'll change, since it echoes in here :)

PR #176, opened 2025-03-21
🌱 Bump golangci-lint to v1.64.7

kashifest:
This PR also bumps golangci-lint-action to 6.5.2. While doing that it removes some of the old unsupported golangci-lint settings.
Signed-off-by: Kashif Khan <kashif.khan@est.tech>

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,5 +1,5 @@
 run:
-  deadline: 10m
+  go: "1.23"
 linters:
   disable-all: true
   enable:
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
/approve

.golangci.yaml
```diff
@@ -119,11 +120,6 @@ issues:
   - linters:
     - revive
     text: "exported: exported method .*\\.(Reconcile|SetupWithManager|SetupWebhookWithManager) should have comment or be unexported"
```
kashifest, line 122: I wonder how these were working before since these rules should go under `exclude-rules` not `skip-files`

dtantsur: I wonder if it's a copy-paste and the exclude is not needed

PR #178, opened 2025-03-21
⚠️ Stop defaulting DHCP rangeBegin/rangeEnd

dtantsur:
The current logic is somewhat arbitrary and may be confusing to
operators. Furthermore, the IP range handling brings a dependency on
the unmaintained netipx library. It is also the only reason we have
a mutating webhook, which otherwise can be removed.

This is a breaking change. Users must provide rangeBegin/rangeEnd.


Rozzii (commented):
/lgtm

PR #183, opened 2025-03-24
🌱 Fix the broken golang-ci linter

dtantsur:
We don't understand yet how it happened, but the current configuration
is outdated and broken, which is now causing a CI breakage.

This change fixes the configuration, fixes the issues that the linter
now detects and adds a configuration validation step to "make lint".

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,8 @@
+version: "2"
 run:
   go: "1.23"
 linters:
-  disable-all: true
+  default: fast
   enable:
   - asasalint
   - asciicheck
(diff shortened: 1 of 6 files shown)


.golangci.yaml
```diff
@@ -1,7 +1,8 @@
+version: "2"
 run:
   go: "1.23"
 linters:
-  disable-all: true
+  default: fast
```
tuminoid, line 5: Hmm, removing disable-all means we don't explictly define the set we want to run. by removing it we went from 32 to 35, while enabling only "importas". Some 2 linters go enabled.

Is there no replacement for the disable-all in the new config?

dtantsur: There is, but we also used `fast: true`? If the intention is to disable all explicitly, then the fast option is unused

dtantsur:
/hold

We might be picking up golangci-lint 2.0

PR #184, opened 2025-03-24
🌱 Enable more linters

dtantsur:
Uncomments some linters and adds some more hand-picked linters that
I found meaningful.

Some more linters are added as commented: these are the ones I find
useful but cannot enable right away.


Rozzii (commented):
/approve

tuminoid (commented):
/lgtm

Good work. We can use this as input to https://github.com/orgs/metal3-io/discussions/2098 as you did do a check whats valuable to us and whats not (at least in your opinion). This list is 68 linters long, which means we have a gap of 20-30 linters across all repos.

Fixes do not seem too massive, but of course we have some of them uncommented right now.

tuminoid:
Duh, needs rebase.

PR #188, opened 2025-03-25
🌱 Functional tests: decrease individual timeouts

dtantsur:
Provisioning Ironic very rarely takes more than 2 minutes, I'm using
5 minutes just to be safe. Waiting 15 minutes is an overkill, at
this point the process is definitely stuck, and we run the risk of
reaching the overall timeout of 60 minutes.

Wait for a failure even less (1.5 minutes).

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -223,7 +223,7 @@ func WaitForIronic(name types.NamespacedName) *metal3api.Ironic {
 
 		logResources(ironic, "")
 		return false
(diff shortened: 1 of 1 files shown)


dtantsur:
> Ran 12 of 12 Specs in 1421.072 seconds

2 minutes per test on average

Rozzii (commented):
/approve

tuminoid (commented):
/lgtm

PR #190, opened 2025-03-25
🐛 Do not clear the installed version when reconciling

dtantsur:
The idea was to signal that reconciliation is in progress, but it's
already clear from requestedVersion != installedVersion. Clearing
this field makes it impossible to know which version we're upgrading
from.


tuminoid (commented):
/lgtm

PR #192, opened 2025-03-25
🐛 Fix readiness detection for updated daemon sets 

dtantsur:
Apparently, NumberUnavailable is not enough for that, we also need
to check the updated number.

Add a test on upgrading an HA environment with a database


lentzi90 (commented):
/lgtm

PR #195, opened 2025-03-26
🌱 Use a multi-node cluster in functional tests

dtantsur:
This PR enables real HA testing on a multi-node minikube cluster.

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -526,6 +526,7 @@ func VerifyIronic(ironic *metal3api.Ironic, assumptions TestAssumptions) {
 	if !assumptions.withHA {
 		currentIronicIPs = getCurrentIronicIPs(ctx, ironic.Namespace, ironic.Name)
 	}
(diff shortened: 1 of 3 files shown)


test/suite_test.go
```diff
@@ -230,6 +230,7 @@ func WaitForIronic(name types.NamespacedName) *metal3api.Ironic {
 
 		writeYAML(ironic, ironic.Namespace, ironic.Name, "ironic")
 		GinkgoWriter.Printf("Current status of Ironic: %+v\n", ironic.Status)
+		logResources(ironic, "")
```
lentzi90, line 233: Do we need to keep this? I'm thinking we could put it in an AfterEach or something so that we get the status when things fail but not all the time.

dtantsur: AfterEach is a good idea. I'll cut this commit out of the PR.

lentzi90 (commented):
/approve

PR #210, opened 2025-03-27
🌱 Dont run golangci-lint separately on internal/controller

dtantsur:
It's not a separate Go module and is linted with the root directory.


Rozzii (commented):
/approve

Issue #196, opened 2025-03-26
Better process of picking up refreshed TLS certificates


dtantsur:
Kubernetes does not restart pods when a mounted secret changes. This poses a problem with TLS certificates: they need regular refreshing. The current solution is based on python-inotify and involves [detecting the changed file and killing the main process of the container](https://github.com/metal3-io/ironic-image/blob/8424709354a75cfa1bb5e472c81259d13af3c6f8/scripts/tls-common.sh#L110-L113), triggering a pod restart.

This procedure is not exactly intuitive or easy to debug. But it poses an even larger issue in the HA context: it will cause all replicates of Ironic to restart simultaneously. To avoid downtime, we need something else.

What if instead IrSO will provide the UID and resourceVersion of each secret in an environment variable? Like this:

```
IRSO_TLS_SECRET_VERSION=80bcd674-62f3-4da5-a06c-3fcc60e5ccee/5186
```

Then every time the secret is re-created (with a different UID) or updated (with a new version), the pod template in the Ironic daemon set will change, triggering a proper rolling update.

lentzi90:
How would IrSO detect the secret change? Or is this only about IrSO provided secrets?
In either case I think it is already a sort of de-facto standard to include the hash of the secret either as suffix of the name or (kustomize) or as a label/annotation in the pod template. You get the same forced rollout but this is a "common" method that would cause less surprises.

PR #212, opened 2025-03-27
🐛 Validate TLS secret and add an owner reference

dtantsur:
Add validation of the TLS secret on the controller level.

Ensure an owner reference on it to receive updates whenever the secret
is updated (will be required for bug https://github.com/metal3-io/ironic-standalone-operator/issues/196).


lentzi90 (commented):
Is the owner ref just to be notified of changes? I'm not sure this is a good idea. If the secret is created by an external tool/user we should not own it.
Would it be possible to set some kind of explicit watch on it instead?

dtantsur:
(Side note: I've split out the commit with refactoring while we're discussing the approach to TLS secrets: https://github.com/metal3-io/ironic-standalone-operator/pull/217)

lentzi90:
> > Would it be possible to set some kind of explicit watch on it instead?
> 
> Well, that's the problem. We can create a watch but we need a way to map a secret back to the Ironic instance. How do we do that without an owner reference?
> 
> I cannot find any prior art for that. I'm not sure if annotations is a good answer.

I see :disappointed: 
It still feels wrong to me to set an owner reference but perhaps it is the best solution.

Thinking out loud here. What if we could solve this in another way in the ironic-image? I have seen that kubernetes adds in some jitter to avoid situations when multiple things might happen simultaneously. For us, this could look something like "oh the certificate changed, I will restart the container at a random time between now and 5 min".

Or perhaps we could even make use of some internal communication? Can one ironic instance check the health/availability of the other in the cluster? Then we could do the restart only if the other instances were healthy.

dtantsur:
> For us, this could look something like "oh the certificate changed, I will restart the container at a random time between now and 5 min".

Could work, but what we do with inotify feels equally wrong (= hacky). At the very least, this stuff will be very difficult to debug if anything goes wrong.

> Can one ironic instance check the health/availability of the other in the cluster?

It does but it's not instantaneous (there are no notifications only periodic polling).

lentzi90 (commented):
Ok it seems like we don't have any better alternatives, so let's take this forward. We can always rework it later if it turns out to be a major issue.
/lgtm

dtantsur:
Sorry, had to rebase, could you re-review?

lentzi90 (commented):
/lgtm

tuminoid (commented):
/approve

PR #219, opened 2025-04-01
🐛 Mount the MariaDB secret when possible

dtantsur:
Passing the password as an environment varible is less secure.
Starting with ironic-image 29.0, a secret mount is accepted.

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -160,8 +160,8 @@ func buildExtraConfigVars(ironic *metal3api.Ironic) []corev1.EnvVar {
 	return result
 }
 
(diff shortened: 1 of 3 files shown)


tuminoid (commented):
Do we have sufficient tests for mounted secrets? Asking as the test files are not updated in this PR.

dtantsur:
> Do we have sufficient tests for mounted secrets? Asking as the test files are not updated in this PR.

If mounting the database secret breaks, the database/HA tests break as well. This is what happened on my first attempt here: https://github.com/metal3-io/ironic-standalone-operator/actions/runs/14198440920

tuminoid (commented):
/lgtm

PR #220, opened 2025-04-01
See how much breaks with ReadOnlyRootFileSystem

dtantsur:


dtantsur:
/cc @Rozzii 

in case you're interested

tuminoid:
> Hmm, this is probably pretty naive: I need to mount some volumes for data, don't I?

Yes, you need to mount `/conf` and `/data` similar to `/shared`.

dtantsur:
Moving over to https://github.com/metal3-io/ironic-standalone-operator/pull/221

PR #221, opened 2025-04-01
✨ Use ReadOnlyRootFileSystem in upgrade jobs

dtantsur:

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -27,8 +27,6 @@ const (
 	authDir   = "/auth"
 	certsDir  = "/certs"
 	sharedDir = "/shared"
-	dataDir   = "/data"
-	confDir   = "/conf"
 
 	knownExistingPath = "/images/ironic-python-agent.kernel"
 )
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -27,6 +27,8 @@ const (
 	authDir   = "/auth"
 	certsDir  = "/certs"
 	sharedDir = "/shared"
+	dataDir   = "/data"
+	confDir   = "/conf"
 
 	knownExistingPath = "/images/ironic-python-agent.kernel"
 )
--- a/pkg/ironic/upgrades.go
+++ b/pkg/ironic/upgrades.go
@@ -71,10 +71,11 @@ func newMigrationTemplate(cctx ControllerContext, ironic *metal3api.Ironic, data
 				Capabilities: &corev1.Capabilities{
 					Drop: []corev1.Capability{"ALL"},
 				},
+				ReadOnlyRootFilesystem: ptr.To(hasReadOnlyRootFilesystem(cctx)),
 			},
 		},
 	}
-	return corev1.PodTemplateSpec{
+	return addDataVolumes(cctx, corev1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels: map[string]string{
 				metal3api.IronicAppLabel:     ironicDeploymentName(ironic),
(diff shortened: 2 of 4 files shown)


tuminoid (commented):
/approve

Looks solid to me.

tuminoid:
/retest

Hmm, it does seem that HA/upgrade has issues with this though.
/approve cancel

dtantsur:
```
+++ mkdir -p /certs/ironic
mkdir: cannot create directory '/certs/ironic': Read-only file system
```

I wonder why we do these in ironic-image if we expect the root filesystem to be read-only..

dtantsur:
```
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
```

It does not feel like this stuff is easy to make work :( I'll probably wait for @Rozzii to return

/cc @Rozzii

tuminoid:
> ```
> 
> FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
> 
> ```
> 
> 
> 
> It does not feel like this stuff is easy to make work :( I'll probably wait for @Rozzii to return
> 
> 
> 
> /cc @Rozzii 

/tmp being writable is kind of implicit requirement, another emptyDor mount should solve it. Alternatively, we make /tmp a symlimk to /data/tmp in Ironic image. 

Lets discuss the options next week.

tuminoid:
> > Lets discuss the options next week.
> 
> Yeah, I agree, let's have a plan before I move on with this PoC. Maybe Adam had different plans. Maybe we can consolidate some of the emptyDirs before their number reaches two digits :)

He kind of consolidated them under /data and /conf, but then we also have /certs, /shared and implicit ones like /tmp etc. 

Technically, number of emptyDirs is irrelevant, they're very lightweight, and I personally argued using individual emptyDirs instead of moving all confs/data under shared dirs, so it is a balancing act between being too custom and dirs being too polluted. I think we could live with 5-6 mounts, if we don't get more surprises.

dtantsur:
> Alternatively, we make /tmp a symlimk to /data/tmp in Ironic image.

Hmm, it looks like Adam has actually tried to do that, I wonder why it did not work.

dtantsur:
```
  File "/usr/lib/python3.9/site-packages/ironic/common/rpc.py", line 22, in <module>
    from ironic.conf import CONF
  File "/usr/lib/python3.9/site-packages/ironic/conf/__init__.py", line 27, in <module>
    from ironic.conf import default
  File "/usr/lib/python3.9/site-packages/ironic/conf/default.py", line 422, in <module>
    default=tempfile.gettempdir(),
  File "/usr/lib64/python3.9/tempfile.py", line 307, in gettempdir
    tempdir = _get_default_tempdir()
  File "/usr/lib64/python3.9/tempfile.py", line 223, in _get_default_tempdir
    raise FileNotFoundError(_errno.ENOENT,
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/tmp', '/var/tmp', '/usr/tmp', '/']
```

Ahhhh, this is curious. Ironic fails when trying to figure out the default value for the `tempdir` option. So even when we override it, a usable global directory must be present.

There are so many places in Ironic that do not respect CONF.tempdir that it's useless anyway :(

dtantsur:
Unfortunately, it's not going to work without something like https://github.com/metal3-io/ironic-image/pull/664

Rozzii (commented):
Yeah I was mounting emptyDir to /var/tmp during my tests and pointing ironic to use this, I am fine with different arrangements also I have added my comments to https://github.com/metal3-io/ironic-image/pull/664 .

dtantsur:
Wow, finally passed the CI \o/

I'll leave it up to you to decide if we want to merge it as it is or wait for the conversation to happen.

tuminoid:
> Wow, finally passed the CI \o/
> 
> I'll leave it up to you to decide if we want to merge it as it is or wait for the conversation to happen.

Keepalived ROFS PR is still open btw. IRSO is not using that for anything?

dtantsur:
> Keepalived ROFS PR is still open btw. IRSO is not using that for anything?

IrSO supports keepalived, but this patch only concerns the upgrade jobs, not the running containers.

Rozzii (commented):
/lgtm

dtantsur:
@elfosardo hey, could you take a look? Would be good to get this out of my backlog.

tuminoid (commented):
/approve
/hold
in case @elfosardo has something.

elfosardo (approved):
/unhold

PR #226, opened 2025-04-02
🌱 Prepare for multi-node functional tests

dtantsur:
This change adapts the testing code for several control plane nodes.
When Kind is not used, the tests now detect which pod is running Ironic.
In the HA scenario, all control plane nodes are assumed to have Ironic.


tuminoid (commented):
/approve

PR #230, opened 2025-04-04
⚠️ Drop support for Ironic 27.0 with HA

dtantsur:
This allows removing handling of the deprecated auth-config from
the API credentials secret. Nobody should be using 27.0 with HA anyway.


tuminoid (commented):
/lgtm

PR #232, opened 2025-04-04
🐛 Do not create conflicting labels on pods and jobs

dtantsur:
Several things got mixed in my mind when writing the upgrade patch.
I intended all resources to have a label mapping them back to Ironic and
its version (for user convenience), but ended up using the label from
the selector of an Ironic deployment/daemonset. Which is clearly a bad
idea.

Create a new label purely for tracking purposes and leave the selector
label only for the pod selector in the deployment/daemonset.


Rozzii (commented):
/lgtm

PR #236, opened 2025-04-07
🌱 Refactoring: group API objects in the new Resources struct

dtantsur:
This PR modifies high-level functions in pkg/ironic to accept a new
Resources structure with Ironic, Database, APISecret and TLSSecret
objects.

The initial thought was to only pass the objects that each function
needs. Unfortunately, higher-level functions need all resources anyway.
Now I need to pass the TLS secret object, which means updating a lot of
functions with a new argument. In this PR, every function that currently
uses the API secret (directly or indirectly) now accepts Resources.

During the refactoring, some redundant arguments have been removed.
E.g. nothing actually used the domain argument passed to some functions.
I'm also removing the assumption that the API secret can be empty (it
cannot, and it has been this way for a while).

### DIFF
--- a/internal/controller/ironic_controller.go
+++ b/internal/controller/ironic_controller.go
@@ -178,14 +178,22 @@ func (r *IronicReconciler) handleIronic(cctx ironic.ControllerContext, ironicCon
 		}
 	}
 
+	var tlsSecret *corev1.Secret
 	if tlsSecretName := ironicConf.Spec.TLS.CertificateName; tlsSecretName != "" {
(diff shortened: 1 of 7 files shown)


Rozzii (commented):
/lgtm

dtantsur:
@tuminoid @Rozzii @lentzi90 could you please approve if you agree? This PR will conflict with everything I'm planning to implement this week.

Rozzii (commented):
/approve

In principle it makes sense.

PR #237, opened 2025-04-07
🐛 Do not requeue while waiting for external events

dtantsur:
The operator is going to be notified via its ownership of Deployments,
DaemonSets and Jobs.


tuminoid (commented):
/lgtm

PR #238, opened 2025-04-07
🌱 Go back to using `minikube image load` with a tarball

dtantsur:
Building the image on all 3 nodes causes CI jobs to run out of disk
space. Unfortunately, the image load command does not work with podman
directly, so we need to use the same trick as with kind: export
a tarball, then upload it to the cluster.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #239, opened 2025-04-08
🐛 Fix detection of a completed Deployment

dtantsur:
Per somewhat hidden documentation [1], we must inspect the Progressing
condition and its Reason. Available is still True directly after an update
because the old replica is still running.

On top of that, stop reconciliation with a fatal error when the
deployment no longer progresses.

Unfortunately, client-go does not expose the necessary Reason constants,
so I have to hardcode them based on the values from the documentation
and real-world testing.

[1] https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

### DIFF
--- a/pkg/ironic/utils.go
+++ b/pkg/ironic/utils.go
@@ -107,24 +107,29 @@ func getDeploymentStatus(cctx ControllerContext, deploy *appsv1.Deployment) (Sta
 		return inProgress("deployment not ready yet")
 	}
 
-	var available bool
-	var err error
+	var available, updated bool
 	for _, cond := range deploy.Status.Conditions {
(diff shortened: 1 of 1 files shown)


dtantsur:
/hold

Even this is not enough.

dtantsur:
/cc @lentzi90 

Your expertise is needed here. This patch is an outcome of trial-and-error. I don't think there are better options but I might very well be missing something.

Rozzii (commented):
/lgtm

lentzi90 (commented):
It looks correct to me. It is quite ridiculous how hard it is to determine if a deployment is in good shape or not...
/approve

PR #241, opened 2025-04-09
🌱 Tests: only collect logs from existing containers and IPs from active pods

dtantsur:
Multinode tests may end up with a few pods that fail to schedule. These
don't have container logs, which causes the logs collection to fail.

Similarly, we end up with empty IP addresses if we collect HostIPs from
such pods. Just ignore them.

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -395,7 +395,9 @@ func getCurrentIronicIPs(ctx context.Context, namespace, name string) []string {
 	addresses := make([]string, 0, len(pods.Items))
 	for _, pod := range pods.Items {
 		// Only use one address per pod, no need for both IP families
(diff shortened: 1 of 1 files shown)


tuminoid (commented):
/approve

tuminoid:
> Multinode tests may end up with a few pods that fail to schedule. These don't have container logs, which causes the logs collection to fail.

But we do keep collecting logs from pods that did run but failed, right?

tuminoid:
> > But we do keep collecting logs from pods that did run but failed, right?
> 
> Yeah, from everything in ContainerStatuses (this includes failed containers).

Can you then update title, as it reads now "only frrom active", which can be a bit misleading IMO.

Rozzii (commented):
/lgtm

PR #244, opened 2025-04-11
🌱 Tests: install cert-manager and mariadb-operator using Helm

dtantsur:
We need Helm for MariaDB Operator, so let's start with using it
for cert-manager too.

Also bump cert-manager to 1.17.1.


Rozzii (commented):
/lgtm

PR #247, opened 2025-04-13
:seedling: Enable unparam linter

winiciusallan:
This PR:
* Enable unparam linter
* Address issues related with unused params in some functions

Fixes: #209

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -67,7 +67,7 @@ linters:
   - tparallel
   - typecheck
   - unconvert
-  #- unparam
+  - unparam
   - unused
   - usestdlibvars
   - usetesting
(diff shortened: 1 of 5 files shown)


winiciusallan:
However, there are some errors that I want to know if reasonable to consider as false positives.

```go
internal/controller/ironic_controller.go:99:102: `(*IronicReconciler).setCondition` - `value` always receives `false` (unparam)

internal/controller/utils.go:47:69: `setCondition` - `status` always receives `metal3api.IronicStatusReady` (`"Ready"`) (unparam)
```

dtantsur:
@winiciusallan to avoid ignoring the error, you can rename `(*IronicReconciler).setCondition` to `setNotReady` and drop the argument. I'm also fine with hardcoding `IronicStatusReady` in utils for now. We don't have any other conditions in the API.

lentzi90 (commented):
/lgtm

PR #249, opened 2025-04-15
🐛 Track TLS certificates using an annotation instead of inotify

dtantsur:
With the inotify approach, all pods in the daemon set will stop roughly
at the same time, causing unavailability. In this approach, a new
annotation is added with the UID and version of the TLS secret. When
it is rotated, the pods undergo a rolling update.

Closes: #196 

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -3,6 +3,7 @@ package ironic
 import (
 	"errors"
 	"fmt"
+	"maps"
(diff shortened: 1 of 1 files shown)


dtantsur:
/cc @lentzi90 

Could you take a look? Would be good to get this out of my backlog.

lentzi90 (commented):
/lgtm

PR #250, opened 2025-04-16
🌱 Separate test on HA with custom TLS and authentication secrets

dtantsur:
TLS is sufficiently different in the HA architecture, so its worth
testing it separately. This will avoid regressions when we'll be working
on RPC TLS.

Throw in a custom authentication secret to increase the matrix.


Rozzii (commented):
/lgtm

PR #254, opened 2025-04-23
✨ Partial TLS support for JSON RPC

dtantsur:
I'm not sure what has happened, but after the TLS+RPC test merged,
the RPC bus switched to using TLS, so the test is now failing.

Instead of trying to fight it, I'm updating the code to properly
support TLS on JSON RPC when TLS is configured for the API.
The plan was to always enable TLS, but this is better than nothing.

Add a knob to disable TLS validation on RPC specifically.


lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #258, opened 2025-04-25
✨ Promote the HighAvailability feature gate to Beta status

dtantsur:
With the recent fixes and CI additions, this feature should be safe
to test by the wider audience. It's still disabled by default.

As a reminder, only virtual media is supported in HA mode for now.

### DIFF
--- a/api/v1alpha1/features.go
+++ b/api/v1alpha1/features.go
@@ -11,7 +11,7 @@ const (
 
 var (
 	availableFeatures = map[featuregate.Feature]featuregate.FeatureSpec{
-		FeatureHighAvailability: {Default: false, PreRelease: featuregate.Alpha},
+		FeatureHighAvailability: {Default: false, PreRelease: featuregate.Beta},
 	}
 
 	CurrentFeatureGate = featuregate.NewFeatureGate()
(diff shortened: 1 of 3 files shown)


tuminoid:
> With the recent fixes and CI additions, this feature should be safe to test by the wider audience. It's still disabled by default.

AFAIK in Kubernetes ecosystem, beta features are enabled by default, and only alpha is disabled. I have no strong opinion either way, just noting it, in case we want to follow the convention. Although, I'm thinking should HA be enabled by default ever as it usually has further requirements.

dtantsur:
> Although, I'm thinking should HA be enabled by default ever as it usually has further requirements.

What do you mean? I think it should be fine to eventually enable. Maybe we can do it right now. After all, the entire IrSO is still a beta :D

tuminoid:
> > Although, I'm thinking should HA be enabled by default ever as it usually has further requirements.
> 
> What do you mean? I think it should be fine to eventually enable. Maybe we can do it right now. After all, the entire IrSO is still a beta :D

I mean, any HA system is always more complex to setup than single instance. We would raise the learning curve by enabling it by default?

dtantsur:
> I mean, any HA system is always more complex to setup than single instance. We would raise the learning curve by enabling it by default?

Oh, hold on, we're talking about different things. I'm talking about enabling the *feature gate* by default, not the feature itself. I.e. right now setting `HighAvailability: true` fails by default because it's gated. I'm considering making it work by default but still as an opt-in. Makes sense?

tuminoid:
> > I mean, any HA system is always more complex to setup than single instance. We would raise the learning curve by enabling it by default?
> 
> Oh, hold on, we're talking about different things. I'm talking about enabling the _feature gate_ by default, not the feature itself. I.e. right now setting `HighAvailability: true` fails by default because it's gated. I'm considering making it work by default but still as an opt-in. Makes sense?

Ah, my bad, makes sense. In retrospect I don't know I even could read it THAT wrong :blush:

tuminoid (commented):
/approve

PR #251, opened 2025-04-16
🌱 Use MariaDB operator in the functional tests

dtantsur:
To replace the deprecated mariadb-image and IronicDatabase API, this
change introduces mariadb-operator into the functional testing.

A single instance is configured during the preparation phase. To ensure
that tests are isolated, a database and a user are created for each test.

Unfortunately, the dependency on mariadb-operator goes the Go version in
tests to be raised and quite a few transient dependencies to be added.

Closes: #197
Signed-off-by: Dmitry Tantsur <dtantsur@protonmail.com>

PR #259, opened 2025-04-25
⚠️ Drop deprecated IronicDatabase API and Ironic.DatabaseName

dtantsur:
These features rely on mariadb-image which does not support upgrades and
cannot be run in an HA mode. Users should use Ironic.Database instead to
link to a properly maintained database (e.g. installed by
mariadb-operator).

Requires:
- https://github.com/metal3-io/ironic-standalone-operator/pull/251


lentzi90 (commented):
/approve

PR #260, opened 2025-04-25
🌱 Tests: repair tests when using Kind

dtantsur:
We should get rid of Kind support eventually, but mariadb-image still
uses it.


tuminoid (commented):
/approve

PR #266, opened 2025-05-21
:book: add CII badge

tuminoid:
Add CII best practices badge. Issues will be filed for those missing criteria that prevent us from getting a passing badge.


lentzi90 (commented):
/lgtm

PR #275, opened 2025-05-30
🌱 Test: shorten the pod status printed to stdout

dtantsur:
Pod status are very long, printing all of them makes the output
unreadable and may overflow disk space on a github worker.

Only print the pod phase and the containers that are not ready.

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -225,7 +225,20 @@ func logResources(ironic *metal3api.Ironic, suffix string) {
 	Expect(err).NotTo(HaveOccurred())
 
 	for _, pod := range pods.Items {
-		GinkgoWriter.Printf("... status of pod %s: %+v\n", pod.Name, pod.Status)
+		var notReady []string
+		if pod.Status.Phase != corev1.PodSucceeded {
+			for _, cont := range pod.Status.InitContainerStatuses {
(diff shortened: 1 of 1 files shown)


test/suite_test.go
```diff
@@ -225,7 +225,20 @@ func logResources(ironic *metal3api.Ironic, suffix string) {
...
+					notReady = append(notReady, cont.Name)
+				}
+			}
+			for _, cont := range pod.Status.ContainerStatuses {
+				if !cont.Ready {
+					notReady = append(notReady, cont.Name)
+				}
+			}
+		}
+		GinkgoWriter.Printf("... status of pod %s: %s, not ready: %+v\n", pod.Name, pod.Status.Phase, notReady)
```
tuminoid, line 241: This is printed even if notReady is empty.

tuminoid: I was thinking of just `if` to skip it if notReady is empty?

dtantsur: I want something to get printed for ready pods. It's especially useful for upgrade jobs - to distinguish between finished jobs and jobs that have never run at all.

tuminoid: I can live with that too.

Rozzii (commented):
/lgtm

PR #279, opened 2025-06-02
🌱 Split the functional test into upgrades and regular tests

dtantsur:
The tests are pretty long, especially the upgrade ones. Splitting them
into a new job will allow for a faster CI and will hopefully avoid
running out of resources in the future.

The patch adds tags "ha" and "upgrade" to all tests where applicable.

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -8,7 +8,6 @@ cd "${REPO_ROOT}/test"
 LOGDIR="${LOGDIR:-/tmp/logs}"
 JUNIT_OUTPUT="${JUNIT_OUTPUT:-${LOGDIR}/report.xml}"
 TEST_TIMEOUT="${TEST_TIMEOUT:-90m}"
-EXTRA_ARGS="${EXTRA_ARGS:-}"
 
 . testing.env
 
--- a/test/run.sh
+++ b/test/run.sh
@@ -8,9 +8,16 @@ cd "${REPO_ROOT}/test"
 LOGDIR="${LOGDIR:-/tmp/logs}"
 JUNIT_OUTPUT="${JUNIT_OUTPUT:-${LOGDIR}/report.xml}"
 TEST_TIMEOUT="${TEST_TIMEOUT:-90m}"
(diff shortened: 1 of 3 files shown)


test/run.sh
```diff
@@ -8,9 +8,16 @@ cd "${REPO_ROOT}/test"
 LOGDIR="${LOGDIR:-/tmp/logs}"
 JUNIT_OUTPUT="${JUNIT_OUTPUT:-${LOGDIR}/report.xml}"
 TEST_TIMEOUT="${TEST_TIMEOUT:-90m}"
+EXTRA_ARGS="${EXTRA_ARGS:-}"
```
tuminoid, line 11: What kind of input are you expecting to pass this variable?

Converting string to array will make input of `EXTRA_ARGS="--ginkgo.label-filter foo"` a single string (quoted) and it will expand on L23 as `' --ginkgo.label-filter foo'` and ginkgo will treat it as a single string and fail as it does not understand it.

You need to parse it, something like this:

```
declare -a EXTRA_ARGS_ARR
if [[ -n "${EXTRA_ARGS:-}" ]]; then
    read -ra EXTRA_ARGS_ARR <<< "${EXTRA_ARGS}"
fi

tuminoid (commented):
Nice improvement!

See the comment about EXRTA_ARGS though.

test/run.sh
dtantsur: Oh, this is a leftover from a previous iteration, I'll drop it.

tuminoid (commented):
/approve

Whats the plan with functional tests in ironic-image and mariadb-image, planning on applying this there too, or is this just to speed up IRSO development where it is more relevant?

dtantsur:
> Whats the plan with functional tests in ironic-image and mariadb-image

ironic-image already skips any upgrade tests (we could and probably should add upgrade tests from previous versions, but it is not done).

mariadb-image needs the job removed since the main branch of IrSO no longer supports it.

Issue #269, opened 2025-05-21
Documentation about API inputs/outputs missing


tuminoid:
> Requirement: The project MUST provide reference documentation that describes the external interface (both input and output) of the software produced by the project. [documentation_interface] 

In other repos, we have `docs/api.md` to fulfill this requirement.

/kind documentation

tuminoid:
> > In other repos, we have docs/api.md to fulfill this requirement.
> 
> We have user guide to explain how to use the software, and we need an auto-generated API reference. If we can somehow generate docs/api.md (at the same time as `make manifests`), great. The manual path that BMO did is a dead-end (and that's why I mostly removed api.md back in the deays).

I'm good with auto-generated API reference and the user guide, it does not have to be api.md in the same format as others. I just noted it as how the other repos are doing it, but like you said they're probably all outdated. Maybe we can do the right thing here and then replicate to other repos. We just need to document it in a nice, accessible way.

dtantsur:
Quick googling yielded https://github.com/fybrik/crdoc which we may want to research

Rozzii:
/triage accepted
I second the use of some automated tool.

PR #281, opened 2025-06-06
📖 Auto-generate human-friendly API documentation from CRDs

dtantsur:
Closes: #269


tuminoid (commented):
Nice! The output is not pretty markdown, but rendered it is perfectly fine :)

dtantsur:
> The output is not pretty markdown

Yeah, I also don't like all this embedded HTML. But it gets the job done, and we can iterate later if/when we have cycles.

tuminoid:
I opened couple issues as reminders. We do have existing api.md in those repos with varying quality and links pointing to external service for CRD documentation, so unlike here, some consideration is needed what to keep/replace/remove.

Rozzii (commented):
/lgtm
This looks awesome!

PR #283, opened 2025-06-11
🌱 Enable tagliatelle linter

### LINKED ISSUE #208: Enable "tagliatelle" linter in golangci-lint
Enable "tagliatelle" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

elfosardo:
Closes #208

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -61,7 +61,7 @@ linters:
   - rowserrcheck
   - staticcheck
   - stylecheck
-  #- tagliatelle
+  - tagliatelle
   - testifylint
   - thelper
   - tparallel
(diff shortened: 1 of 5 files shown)


tuminoid (commented):
As discussed in https://github.com/metal3-io/baremetal-operator/pull/2436, this would be an API change. Since this is IRSO and nothing is stable yet, it should be fine, but I would label it as breaking change.

We can also discuss if CIDR should be ever be written in lower case. In BMO we added few similar terms like RAID to ignore list.

/cc @dtantsur

tuminoid (commented):
/approve

Apparently CIDR is hard to type right :D 
Thanks for fixing the typos.

Rozzii (commented):
/lgtm

PR #288, opened 2025-06-17
:seedling: clean up obsolete release variables from Makefile

tuminoid:
We're actually not using these variables with the new Go based release note generator, the logic is all built-in to the binary.


lentzi90 (commented):
/lgtm

PR #302, opened 2025-07-03
🌱 Tests: better logs collection for upgrade debugging

dtantsur:
Something strange is happening in the upgrade jobs, and the current logs
don't provide enough clue, so:
- collect logs on every iteration instead of only at the end
- collect ReplicaSets as well
- sort Events by CreationTimestamp to make the result readable
- collect more resources from each namespace
- collect pods from kube-system


Rozzii (commented):
/lgtm

PR #317, opened 2025-08-06
🌱 Publish the installation manifest on releasing

dtantsur:
This change simplifies installation of future releases by providing an
install.yaml that users can download and `kubectl apply`.


tuminoid (commented):
Tested the manifest, it works nicely after installing cert-manager first.

Otherwise, the makefile logic is from CAPM3 and works as well.

/approve

PR #329, opened 2025-08-11
🌱 Tests: increase the timeout in upgrade tests

dtantsur:
It seems that the HA tests sometime just slightly exceed the current 5
minutes. It's expected that HA upgrades take a bit longer than an
installation since the daemon set updates replicas sequentially.


tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #335, opened 2025-08-14
:seedling: Install kustomize for release workflow

elfosardo:
We use kustomize in the release workflow so we need to install it

Signed-off-by: Riccardo Pittau <elfosardo@gmail.com>


Makefile
```diff
@@ -330,6 +330,7 @@ release-notes: $(RELEASE_NOTES_DIR) $(RELEASE_NOTES)
 
 .PHONY: release
 release:
+	$(MAKE) kustomize
```
dtantsur, line 333: The usual way to do that is

`release: kustomize`

tuminoid (commented):
/approve

PR #349, opened 2025-08-20
:seedling: Tests: increase the node testing timeout for the non-HA case

dtantsur:
Since the recent Ironic changes, creating and deleting a lot of nodes is
taking much more time (~4 minutes instead of 1-2). While we're
investigating the root cause, bump the timeout to unblock the CI.

Add a non-HA test with a database to be able to see if the difference is
because of HA specifically or because of a real database.


lentzi90 (commented):
/lgtm

PR #361, opened 2025-08-29
✨ Add support for overrides for generated pods

dtantsur:
This change adds a new field Overrides to Ironic resource. This field
will allow advanced operators and downstream consumers to inject
containers, annotations, labels and environment variables into the
generated deployments/daemon sets.

The initial patch implements only annotations and labels.

A new feature gate Overrides is added to protect this feature.

### DIFF
--- a/api/v1alpha1/ironic_types.go
+++ b/api/v1alpha1/ironic_types.go
@@ -261,11 +261,11 @@ type Database struct {
 }
 
 type Overrides struct {
-	// Extra annotations to add to each container.
+	// Extra annotations to add to each pod (including upgrade jobs).
 	// +optional
 	Annotations map[string]string `json:"annotations,omitempty"`
 
-	// Extra labels to add to each container.
+	// Extra labels to add to each pod (including upgrade jobs).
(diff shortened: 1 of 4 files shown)


dtantsur:
Hmm, should this apply to upgrade jobs as well?

/hold

lentzi90 (commented):
I agree something like this will be needed but this feels a bit inconsistent to me. Can we not put it directly in the spec (like `nodeSelector`) instead of under `overrides`?
If the idea is to have more fields for extra containers, volumes, etc later it would be quite strange that `nodeSelector` is separate from all this.

Labels and annotations specifically could perhaps be under an additional field though to make it more obvious what they do. I looked around a bit and found that in [prometheus](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusSpec), they do it as `spec.podMetadata`. Maybe we can copy that?
No particularly strong opinions on my side regarding this though so feel free to talk me out of it 🙂

Rozzii (commented):
/lgtm

PR #362, opened 2025-09-01
🌱 Export IPA_BASEURI to have stable tests

smoshiur1237:
We have seen flaky test where it was taking long time to download IPA images and causing failure in the test. The addition of the IPA_BASEURI var in the test could fix the flaky behavior.

Failed Test: We can check the ramdix downloader file in logs. We can see it was timedout after 30 minutes and trying to download from opendev link.
https://github.com/metal3-io/ironic-standalone-operator/actions/runs/17063752211/job/48478369021

### DIFF
--- a/config/testing/kustomization.yaml
+++ b/config/testing/kustomization.yaml
@@ -2,4 +2,4 @@ resources:
 - ../default
 
 patches:
-- path: manager_feature_gates_patch.yaml
+- path: manager_irso_config_patch.yaml
--- a/config/testing/manager_irso_config_patch.yaml
+++ b/config/testing/manager_irso_config_patch.yaml
@@ -11,3 +11,5 @@ spec:
           env:
             - name: FEATURE_GATES
               value: HighAvailability=true,Overrides=true
+            - name: IPA_BASEURI
+              value: https://artifactory.nordix.org/artifactory/openstack-remote-cache/ironic-python-agent/dib
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"maps"
 	"net/netip"
+	"os"
 	"strconv"
 	"strings"
 
@@ -558,6 +559,8 @@ func newIronicPodTemplate(cctx ControllerContext, resources Resources) (corev1.P
 	var ipaDownloaderVars []corev1.EnvVar
 	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
 		"IPA_BRANCH", cctx.VersionInfo.AgentBranch)
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"IPA_BASEURI", os.Getenv("IPA_BASEURI"))
 
 	volumes, mounts := buildIronicVolumesAndMounts(cctx, resources)
 	sharedVolumeMount := mounts[0]


tuminoid (commented):
In `pkg/ironic/containers.go`, `newIronicPodTemplate` add something like:

```go
if resources.Ironic.Spec.DeployRamdisk.IpaBaseURI != "" {
    ipaDownloaderVars = append(ipaDownloaderVars, corev1.EnvVar{
        Name:  "IPA_BASEURI",
        Value: resources.Ironic.Spec.DeployRamdisk.IpaBaseURI,
    })
}
```

api/v1alpha1/ironic_types.go
```diff
@@ -181,6 +181,10 @@ type DeployRamdisk struct {
 	// SSHKey is the contents of the public key to inject into the ramdisk for debugging purposes.
 	// +optional
 	SSHKey string `json:"sshKey,omitempty"`
+
+	// Dowonloader will use the IpaBaseURI to download the IPA image
+	// +optional
+	IpaBaseURI string `json:"ipaBaseURI,omitempty"`
```
dtantsur, line 187: We don't use the term "IPA" in the fields here (and writing it as "Ipa" is incorrect anyway). Probably call it `baseDownloadURI`.

api/v1alpha1/ironic_types.go
```diff
@@ -181,6 +181,10 @@ type DeployRamdisk struct {
 	// SSHKey is the contents of the public key to inject into the ramdisk for debugging purposes.
 	// +optional
 	SSHKey string `json:"sshKey,omitempty"`
+
+	// Dowonloader will use the IpaBaseURI to download the IPA image
```
dtantsur, line 185: This is not sufficient to use this field: how do users know what to put here? At least an example is needed. Also, how is it going to work with the multi-arch support when we need to invoke the downloader twice?

test/helpers/resources.go
```diff
@@ -20,6 +20,7 @@ var (
 	customImageVersion       string
 	customIPADownloaderImage string
 	customKeepalivedImage    string
+	ipaBaseURI               string
```
dtantsur, line 23: same re consistency and not using "ipa" without context

api/v1alpha1/ironic_types.go
dtantsur: Please also keep the fields sorted

test/testing.env
```diff
@@ -5,6 +5,7 @@ export PROVISIONING_CIDR="172.22.0.1/24"
 export PROVISIONING_RANGE_BEGIN="172.22.0.10"
 export PROVISIONING_RANGE_END="172.22.0.100"
 export PROVISIONING_INTERFACE="eth0"
+export IPA_BASEURI="https://artifactory.nordix.org/artifactory/openstack-remote-cache/ironic-python-agent/dib"
```
dtantsur, line 8: By the way, you don't need to implement the API addition to support this. If you only support the environment variable, it will solve some of the concerns (although we still need to think what to do about the multi-arch situation eventually).

dtantsur (changes requested):

smoshiur1237:
> In `pkg/ironic/containers.go`, `newIronicPodTemplate` add something like:
> 
> ```go
> if resources.Ironic.Spec.DeployRamdisk.IpaBaseURI != "" {
>     ipaDownloaderVars = append(ipaDownloaderVars, corev1.EnvVar{
>         Name:  "IPA_BASEURI",
>         Value: resources.Ironic.Spec.DeployRamdisk.IpaBaseURI,
>     })
> }
> ```
I think the updated changes will pass the IPA_BASEURI var to IPAdownloader. Would you please check?

dtantsur (dismissed):
This is fine for now. Once we get to the multi-arch topic, we may be forced to revert this change though, unless the mirror will get multi-arch support as well.

Rozzii (commented):
/lgtm

Rozzii:
> > /lgtm
> 
> /lgtm cancel Please don't drive-by lgtm when its clearly being worked on still.

Wasn't a drive by, I just missed the last comments, as I had a review window open from earlier time than your comments.

tuminoid (commented):
/lgtm
This works, tested manually and seen in workflow logs.

/hold
@dtantsur PTAL if you're happy with this approach and unhold.

config/testing/manager_feature_gates_patch.yaml
```diff
@@ -11,3 +11,5 @@ spec:
           env:
             - name: FEATURE_GATES
               value: HighAvailability=true
+            - name: IPA_BASEURI
```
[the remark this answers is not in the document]
tuminoid: This was discussed in community meeting: let's rename the file to not mention feature gates as it now contains more configuration, and rebase on top of main as there has been modifications to this file, and then we can merge it.

EXCEPT, we found out the Nordix cache is holding stale copy of IPA, so we need to figure out why it is not updated.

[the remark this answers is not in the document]
tuminoid: Also, we cannot cherry-pick this anymore, since the new feature flag has been introduced that conflicts. We need manual cherry-picks.

Issue #245, opened 2025-04-11
Make the JSON RPC port configurable


dtantsur:
The default port of 8089 may conflict with other workloads. At the very least, we need to be able to change it similarly to other ports (see the `Networking` structure).

There is no suitable option in ironic-image, so use `OS_JSON_RPC__PORT` to pass it.

/kind feature
/help

winiciusallan:
Hi @dtantsur! The user can not already pass it with extraConfig?

```yaml
spec:
  extraConfig:
  - group: json_rpc
    name: port
    value: "8090"
```

Rozzii:
@dtantsur  will take a look, we have agreed that I won't accept this Issue for now.

dtantsur:
> The user can not already pass it with extraConfig?

True. It will change, however, once we start supporting TLS on RPC. If we do it the usual path - using a reverse proxy - we need to pass the same value to httpd and ironic.

I guess this request is a bit premature. Do we have a status "on hold" for bugs?

Rozzii:
I guess the closest to a feature that would become needed later is the the "long-term priority label" . So I have accepted this now but with the long term label.

PR #363, opened 2025-09-03
⚠️ Move the JSON RPC port from 8089 to 6189

dtantsur:
The port 8089 conflicts with kubernetes-nmstate, and is generally quite
risky because of the 808* range. We have chosen a new port from the same
618* range we use for httpd.

A quick internet search revealed no known users of port 6189. A query to
Gemini revealed a malware that was used around 2010 but has all but
disappeared by now.

A new API is added to modify the port in case of future conflicts.

See-also: https://github.com/metal3-io/ironic-image/pull/741
Closes: #245 


tuminoid (commented):
/approve

PR #368, opened 2025-09-05
🌱 Tests: collect logs from cert-manager

dtantsur:
Its installation occasionally times out, it will be good to know why.


Rozzii (commented):
/lgtm

PR #376, opened 2025-09-15
:seedling: Fix OSV-Scanner

erjavaskivuori:
**What this PR does / why we need it**: Adopt reusable workflow from project-infra to run OSV scanner.

Renamed the workflow with `scheduled` prefix as it is a plan to add a workflow for all PRs too.


erjavaskivuori:
/hold

I need to fix the way config.toml is passed for the scanner.

tuminoid:
/hold
Still not working, let's keep all of these on hold, and iterate on one to get the right combination of flags. We'll do that iteration in IPAM PR.

tuminoid (commented):
/approve

Once the project-infra PR merges, we can test this. Until then, keeping the hold.

lentzi90 (commented):
/lgtm

PR #380, opened 2025-09-19
⚠️ Drop support for Ironic older than 30.0

dtantsur:
27.0 and 28.0 are EOL per Ironic policy. 29.0 reaches 6 months tomorrow.
It will be supported for a while because it's a part of the coordinated
release, but from our perspective it's time to drop it as well.
Removing 29.0 in particular allows simplifying a lot of currently
version-dependent logic.

With the upcoming 32.0, IrSO 0.6 will support 3 Ironic versions.


tuminoid (commented):
/lgtm

PR #388, opened 2025-09-30
:sparkles: Switch to Ironic 32.0 by default, prepare release-0.6

elfosardo:
Signed-off-by: Riccardo Pittau <elfosardo@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,7 +73,7 @@ endif
 
 # Image URL to use all building/pushing image targets
 IMG_NO_TAG ?= quay.io/metal3-io/ironic-standalone-operator
-IMG ?= $(IMG_NO_TAG):latest
+IMG ?= $(IMG_NO_TAG):release-0.6
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))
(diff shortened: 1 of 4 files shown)


elfosardo:
this is the update on main https://github.com/metal3-io/ironic-standalone-operator/pull/390

tuminoid:
> @dtantsur I just realized that there are discrepancies in main, specifically the new version of ironic has been added, but of course not the new release we still need to update https://github.com/metal3-io/ironic-standalone-operator/blob/main/docs/releasing.md this is still on hold, I will update main with what needs to be updated before merging this

I think the IRSO process has gone bit off track versus the release process here. The branch etc should be created by the release automation, for example 0.6.0-beta.0 (beta cuts branch, alpha doesn't). Then you implement the CI support for the branch etc. Then you merge stuff there to cut the stable.  As mentioned the 32.0 support should've also  been in main before that.

Do we have some shortcomings in the release docs related to the workflow, or is it not working for IRSO?

tuminoid:
> @tuminoid simply put, the release docs for IrSO is outdated we need to add at least a mention to verify support for recent ironic-image releases, the rest kind of work, just need a refresh

Maybe we can amend verify-release.sh to check for these as well.

PR #406, opened 2025-10-15
🌱 Do not mount service account token on Ironic pods

lentzi90:

Ironic has no need to authenticate or communicate with the Kubernetes API, so there is no reason to mount the service account token. This commit explicitly configures Kubernetes to not mount it. By default, each pod will always mount the default service account if nothing is specified.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #


tuminoid (commented):
/approve

PR #407, opened 2025-10-18
✨ Overrides: allow appending/changing containers and initContainers

dtantsur:
To fascilitate downstream modifications, we need to support appending
more containers to the deployment. For example, OpenShift uses sidecar
containers to manage networking. Unless we want to hardcode all possible
cases, we need to allow overrides for containers.

This change adds two more fields to Overrides: containers and
initContainers. By default, items are appended to the respective lists.
If a name of an existing container is used, the existing container is
replaced instead.

The downside of the approach here is that both the manifests and the API
documentation become huge with the addition of the entire Container
structure from Kubernetes.

Assisted-By: Claude Code (commercial license)


Rozzii (commented):
/approve

Issue #282, opened 2025-06-10
[RFE] Support for ironic-prometheus-exporter


dtantsur:
Ironic is capable of exposing hardware metrics for prometheus. Things like overall health or specific readings can be exposed. IrSO needs to be able to configure Ironic to publish the metrics and install ironic-prometheus-exporter (IPE) alongside Ironic.

The way IPE works now is by reading JSON files from a shared directory where Ironic puts them. There is a desire to rework this architecture (e.g. my merging IPE into Ironic proper), but so far this is how it works. Ironic Image already supports IPE, but we're getting reports that the support is incomplete.

Ideally, I'd prefer to see a new CRD for IPE. Since IPE is tightly bundled with Ironic (i.e. will live in the same pod), it may not make a lot of practical sense though. So it's probably going to be a new configuration section in the Ironic CRD, for instance

```yaml
spec:
  networking:
    prometheusExporterPort: 3306  # or whatever the good default is given host networking
  prometheusExporter:
    enabled: true
```

Over time, more options could be added, the ones I have in mind are

```yaml
spec:
  prometheusExporter:
    enabled: true  # enable the feature
    disableDefaultRules: false  # do not create the default prometheus rules
    defaultCollectionInterval: 30  # how often to poll BMC's by default
```

The sticking point is HA support. The way prometheus is configured, it may not understand a service that is backed by several pods that provide *different* information. It's possible that we'll need something (a new deployment of a new script) that aggregates information from all Ironic instances.

PR #418, opened 2025-11-05
✨ Add support for ironic-prometheus-exporter

dtantsur:
This change exposes the ability to run a prometheus exporter for Ironic
nodes introduced in ironic-image 31.0.

The exporter is configured through the new PrometheusExporter field in
the Ironic spec, which controls:
- Enabling/disabling the exporter container
- Sensor data collection interval
- Metrics endpoint port configuration

When enabled, this deploys the ironic-prometheus-exporter container
alongside the main Ironic containers and configures Ironic to collect
sensor data from BMCs at the specified interval.

If ServiceMonitor resources from prometheus-operator are available,
a ServiceMonitor can be created automatically.

Closes: #282 
Assisted-By: Claude Code (commercial license)

### DIFF
--- a/pkg/ironic/servicemonitor.go
+++ b/pkg/ironic/servicemonitor.go
@@ -2,13 +2,11 @@ package ironic
 
 import (
 	"context"
-	"reflect"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	metal3api "github.com/metal3-io/ironic-standalone-operator/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


pkg/ironic/containers.go
```diff
@@ -549,6 +568,39 @@ func newKeepalivedContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) c
...
+			},
+		},
+		SecurityContext: &corev1.SecurityContext{
+			RunAsUser:  ptr.To(ironicUser),
+			RunAsGroup: ptr.To(ironicGroup),
+			Capabilities: &corev1.Capabilities{
+				Drop: []corev1.Capability{"ALL"},
+			},
+		},
+	}
```
dtantsur: Note to myself: need to mount `/shared`

api/v1alpha1/ironic_types.go
```diff
@@ -295,6 +302,30 @@ type Overrides struct {
...
+// PrometheusExporter defines configuration for Prometheus metrics export.
+type PrometheusExporter struct {
+	// DisableServiceMonitor controls whether a ServiceMonitor resource is created.
+	// Set to true if your cluster does not have prometheus-operator installed,
+	// or when you want to run the exporter but manage Prometheus configuration manually.
+	//
+	// Must be set to true for a highly available deployment. In this case, every replica
+	// provides different metrics, which must be aggregated on the consumer side.
+	// +optional
+	DisableServiceMonitor bool `json:"disableServiceMonitor,omitempty"`
```
pixelsoccupied: Important to note that without a correctly configured service monitor...IPE is basically useless! given that you already have the useful `clusterHasCRD` func...should we actually allow user to delete this and instead maybe update status to note the limitation?

pkg/ironic/servicemonitor.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if err != nil {
+		if client.IgnoreNotFound(err) == nil {
+			// ServiceMonitor doesn't exist, create it
+			cctx.Logger.Info("creating ServiceMonitor", "ServiceMonitor", sm.Name)
+			err = cctx.Client.Create(cctx.Context, sm)
+			if err != nil {
+				return transientError(err)
+			}
+			return updated()
```
pixelsoccupied: In both Creation and Update we are returning "updated"...that's fine?

pixelsoccupied (commented):
good stuff! thanks

pkg/ironic/version.go
```diff
@@ -110,3 +112,16 @@ func (versionInfo VersionInfo) WithIronicOverrides(ironic *metal3api.Ironic) (Ve
 
 	return versionInfo, nil
 }
+
+func checkVersion(resources Resources, version metal3api.Version) error {
+	if resources.BMCCASecret != nil && version.Compare(versionBMCCA) < 0 {
+		return errors.New("using tls.bmcCAName is only possible for Ironic 32.0 or newer")
```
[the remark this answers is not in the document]
dtantsur: Errorf expects an internal error to wrap (the %w argument). Here it's a new error, there is nothing to wrap.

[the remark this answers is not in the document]
pixelsoccupied: i was under impression that error.new is mostly reserved for sentinel errors and basically everywhere we use fmt (regardless of %w use). but this works too! thanks

PR #420, opened 2025-11-06
✨ Add support for custom CA certificates

alegacy:
Adds a new TLS.TrustedCAName field that allows users to provide a ConfigMap containing CA certificates for validating HTTPS connections to image servers and other services. This enables Ironic to access images hosted behind self-signed or custom CA certificates.

The implementation uses Ironic's native WEBSERVER_CACERT_FILE environment variable which configures the webserver_verify_ca option in ironic.conf. This approach is superior to using generic Python environment variables (REQUESTS_CA_BUNDLE, SSL_CERT_FILE) which would override the entire Python requests library trust store globally and could interfere with BMC CA validation and other TLS operations.

The ConfigMap is mounted at /certs/ca/trusted (following existing ironic-image conventions) rather than OS-specific paths like /etc/pki/ca-trust/extracted/pem, making the implementation OS-agnostic. The operator automatically uses the first key found in the ConfigMap, logging warnings for any additional keys that are ignored. This provides flexibility for users while maintaining simplicity.

The operator does not manage or claim ownership of the user-provided ConfigMap, it simply references it for mounting into the Ironic pods.

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -173,7 +173,36 @@ func databaseClientEnvVars(db *metal3api.Database) []corev1.EnvVar {
 	return envVars
 }
 
-func buildIronicEnvVars(resources Resources) []corev1.EnvVar {
+func buildTrustedCAEnvVars(cctx ControllerContext, configMap *corev1.ConfigMap) []corev1.EnvVar {
+	if configMap == nil || len(configMap.Data) == 0 {
+		return nil
+	}
+
+	// Select the first key from the ConfigMap and collect ignored keys
+	var firstKey string
+	first := true
+	for key := range configMap.Data {
+		if first {
+			firstKey = key
+			first = false
+		} else {
+			cctx.Logger.Info("ignoring duplicate key in Trusted CA ConfigMap",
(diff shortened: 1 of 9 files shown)


pkg/ironic/containers.go
```diff
@@ -173,6 +173,41 @@ func databaseClientEnvVars(db *metal3api.Database) []corev1.EnvVar {
...
+	// Select the first key from the ConfigMap and collect ignored keys
+	var firstKey string
+	var ignoredKeys []string
+	first := true
+	for key := range configMap.Data {
+		if first {
+			firstKey = key
+			first = false
+		} else {
+			ignoredKeys = append(ignoredKeys, key)
```
dtantsur: You can issue the warning here and avoid building the ignoredKeys slice

dtantsur:
/approve
/cc @Rozzii 

Adam, I don't think this feature conflicts with your TLS work, but please take a look just in case.

tuminoid:
Approved workflow runs, let's see the tests.

Rozzii:
Looking into this now.

Issue #423, opened 2025-11-14
Ironic deployment rolled out when first BMH is created


lentzi90:
I have noticed a strange behavior. I deploy Ironic using IrSO, wait for everything to be healthy and stable, then I create a BMH.
As soon as I do this, I can see the Ironic pod go into `terminating` state. A new pod then comes up and things start to work again.
I don't see why there is a rollout, but I can see on the Deployment that something is being done. It goes through a rollout, first scaling down the old replicaset and then scaling up the new.

Before creating the BMH:

```console
❯ k get pods -A
NAMESPACE                           NAME                                                             READY   STATUS    RESTARTS   AGE
baremetal-operator-system           baremetal-operator-controller-manager-78b584d66f-gwcr5           1/1     Running   0          4m16s
baremetal-operator-system           ironic-service-5d656b9959-485rm                                  5/5     Running   0          4m18s
```

After:

```console
❯ k get pods -A
NAMESPACE                           NAME                                                             READY   STATUS        RESTARTS   AGE
baremetal-operator-system           baremetal-operator-controller-manager-78b584d66f-gwcr5           1/1     Running       0          4m38s
baremetal-operator-system           ironic-service-5d656b9959-485rm                                  5/5     Terminating   0          4m40s
❯ k -n baremetal-operator-system describe deployments.apps ironic-service
...
OldReplicaSets:  ironic-service-5d656b9959 (0/0 replicas created)
NewReplicaSet:   ironic-service-986bcd77c (1/1 replicas created)
Events:
  Type    Reason             Age    From                   Message
  ----    ------             ----   ----                   -------
  Normal  ScalingReplicaSet  8m49s  deployment-controller  Scaled up replica set ironic-service-5d656b9959 from 0 to 1
  Normal  ScalingReplicaSet  4m17s  deployment-controller  Scaled down replica set ironic-service-5d656b9959 from 1 to 0
  Normal  ScalingReplicaSet  3m46s  deployment-controller  Scaled up replica set ironic-service-986bcd77c from 0 to 1
```

What could be going on to cause this?

Logs from IrSO:

```
2025-11-14T09:17:19Z	INFO	controllers.Ironic	starting reconcile	{"Ironic": {"name":"ironic","namespace":"baremetal-operator-system"}}
[...]

PR #424, opened 2025-11-14
⚠️ Track secret using their content instead of their version

dtantsur:
The current approach of using ResourceVersion means that a new rollout
happens even if only the metadata of a secret is modified. This should
not happen.

Use a fast hash of all data fields instead.

This PR is a breaking change because it will cause all Ironic deployments
to be updated on IrSO upgrade.

Closes: #423 
Generated-By: Claude Code (commercial license)

### DIFF
--- a/pkg/ironic/secrets_test.go
+++ b/pkg/ironic/secrets_test.go
@@ -183,3 +183,90 @@ func TestGenerateSecret(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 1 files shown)


dtantsur:
/cc @lentzi90 @elfosardo 

What do you think?

lentzi90 (commented):
Yeah this makes sense to me. A test would probably be good though.

tuminoid (commented):
/lgtm

PR #437, opened 2025-11-26
:seedling: add missing markdownlint config file and fix issues

tuminoid:
For whatever reason, we have forgotten to bring in the markdownlint config file. This has led to some files having now linting failures, which are also fixed in this PR.


lentzi90 (commented):
/lgtm

PR #441, opened 2025-11-28
:sparkles: use SecretManager to handle access to cluster secrets

### LINKED ISSUE #425: IrSO watches all secrets in the cluster
Because of how the cached client works, IrSO currently maintains an in-memory cache of all secrets in the cluster. We need to do something like BMO's [secret manager](https://github.com/metal3-io/baremetal-operator/blob/main/pkg/secretutils/secret_manager.go) to avoid that.

Maybe consider doing the same for other expensive resources.

/kind bug
/help

tuminoid:

Use the SecretManager from BMO to also handle secrets in IRSO. Instead of holding all cluster secrets in cluster client cache, we only keep the ones that match our labels. This lowers memory consumption but also prevents a risk that the cluster secrets might get dumped. This has been adapted from BMO's SecretManager.

Fixes #425

**Checklist:**

- [x] E2E tests have been added, if necessary.

### DIFF
--- a/pkg/secretutils/label_test.go
+++ b/pkg/secretutils/label_test.go
@@ -1,0 +1,23 @@
+package secretutils
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestAddSecretSelector_NilInput(t *testing.T) {
+	result := AddSecretSelector(nil)
+	assert.Len(t, result, 1)
+}
+
+func TestAddSecretSelector_ExistingMap(t *testing.T) {
+	existing := make(map[client.Object]cache.ByObject)
+	result := AddSecretSelector(existing)
+
+	assert.Len(t, result, 1)
+	// Verify it returns the same map reference (modified in place)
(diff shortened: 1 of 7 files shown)


matthewei (commented):
Do we still need to monitor secret?

pkg/secretutils/label_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	assert.NotNil(t, result)
+	assert.Len(t, result, 1)
+}
+
+func TestAddSecretSelector_ExistingMap(t *testing.T) {
+	existing := make(map[client.Object]cache.ByObject)
+	result := AddSecretSelector(existing)
+
+	assert.NotNil(t, result)
```
dtantsur: nit: duplicated by the Len check

pkg/secretutils/label_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestAddSecretSelector_NilInput(t *testing.T) {
+	result := AddSecretSelector(nil)
+
+	assert.NotNil(t, result)
```
dtantsur: nit: duplicated by the Len check

PR #461, opened 2025-12-09
:warning: Remove support for ironic-image v30

elfosardo:
Also fix support for v33 in some tests


dtantsur:
/approve

Bonus for removing https://github.com/metal3-io/ironic-standalone-operator/blob/main/pkg/ironic/version.go#L22 and any logic associated with it

tuminoid (commented):
/lgtm

Issue #431, opened 2025-11-23
Support ConfigMaps and Secrets for CA


dtantsur:
It looks like cert-manager creates CA as secrets, while in OpenShift CA are usually stored in ConfigMaps (because they're not really secret). I can only suggest supporting both.

/kind feature

PR #463, opened 2025-12-09
🐛 Avoid caching all ConfigMaps in the cluster

dtantsur:
Generalize SecretManager to also handle ConfigMaps. Required for #431.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,6 +8,14 @@ rules:
   - ""
   resources:
   - configmaps
+  verbs:
+  - get
+  - list
+  - update
+  - watch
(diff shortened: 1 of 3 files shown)


dtantsur:
> I think this shouldn't be titled bug fix as it just expands support into configmaps?

It does stop IrSO from listening to all ConfigMaps, although admittedly it's not clear from the summary

tuminoid:
> > I think this shouldn't be titled bug fix as it just expands support into configmaps?
> 
> It does stop IrSO from listening to all ConfigMaps, although admittedly it's not clear from the summary

Ahhh. OK, then it makes sense. Maybe fix the desc? :)

PR #476, opened 2025-12-11
:seedling: update dependabot config for 0.6 and 0.7

tuminoid:
Remove 0.4 and 0.5 and replace with 0.6 and 0.7.

Apparently 0.6 was not updated at all recently ...


tuminoid:
/hold
Merge after 0.7.0 is out to avoid creating PRs into the branch before that.

PR #489, opened 2025-12-12
:sparkles: pass proxy environment variables to ramdisk-downloader

### LINKED ISSUE #308: Pass environment variables to ramdisk-downloader
Hi, 

I'm looking at the CRD spec and as far as I can tell there is no way to pass other environment variables other than IPA_BRANCH. I need to pass HTTPS_PROXY and NO_PROXY to the ramdisk downloader in order to be able to download the ipa-client images. 

Would that be possible?

Thank you!

terror96:
Add support to pass HTTPS_PROXY, HTTP_PROXY and NO_PROXY environment variables to the ramdisk-downloader init container. These environment variables map to IPA_HTTPS_PROXY, IPA_HTTP_PROXY and IPA_NO_PROXY customization values, respectively.

Proxy environment variables are required to enable downloading when routing requests through a designated intermediary server instead of directly to the Internet.

Fixes #308 

**Checklist:**

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -668,6 +668,12 @@ func newIronicPodTemplate(cctx ControllerContext, resources Resources) (corev1.P
 		"IPA_BRANCH", cctx.VersionInfo.AgentBranch)
 	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
 		"IPA_BASEURI", os.Getenv("IPA_BASEURI"))
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"HTTPS_PROXY", os.Getenv("IPA_HTTPS_PROXY"))
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"HTTP_PROXY", os.Getenv("IPA_HTTP_PROXY"))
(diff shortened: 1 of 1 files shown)


pkg/ironic/containers.go
```diff
@@ -668,6 +668,12 @@ func newIronicPodTemplate(cctx ControllerContext, resources Resources) (corev1.P
 		"IPA_BRANCH", cctx.VersionInfo.AgentBranch)
 	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
 		"IPA_BASEURI", os.Getenv("IPA_BASEURI"))
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"HTTPS_PROXY", os.Getenv("IPA_HTTPS_PROXY"))
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"HTTP_PROXY", os.Getenv("IPA_HTTP_PROXY"))
+	ipaDownloaderVars = appendStringEnv(ipaDownloaderVars,
+		"NO_PROXY", os.Getenv("IPA_NO_PROXY"))
```
dtantsur: Why prefix these with `IPA_`, do you expect proxy settings to be IPA-specific?

Ideally, we also need API for ironic-wide proxy settings, but it can wait.

PR #496, opened 2025-12-16
🌱 Enable nakedret linter for golangcilint

### LINKED ISSUE #203: Enable "nakedret" linter in golangci-lint
Enable "nakedret" linter in golangci-lint. Fix any warnings enabling it causes.

/area ci
/help
/triage accepted

smoshiur1237:
Fixes #203 
Description in issue. Enabling and fixing the "nakedret" linter issues

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -43,7 +43,7 @@ linters:
   - mirror
   - misspell
   #- mnd
-  #- nakedret
+  - nakedret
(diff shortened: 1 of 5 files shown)


dtantsur:
If we think that nakedret is potentially useful, maybe actually fix the issues then? I don't think they'll hurt readability.

PR #513, opened 2026-01-19
🌱Uplift k8s.io/utils to match the parent gomod

smoshiur1237:
Uplift gomod in test to match with the parent gomod. This is necessary before doing a patch release.


Rozzii (commented):
/approve

PR #536, opened 2026-02-05
:seedling: Update MariaDB operator installation with retry logic and github repos

elfosardo:
Update the MariaDB operator installation process in test/prepare.sh to improve reliability:
- Change repository URL to the official GitHub repository
- Add helm repo update command to ensure latest chart metadata
- Implement retry logic (5 attempts with 10s delays) for both CRD and operator installations to handle transient failures

### DIFF
--- a/test/prepare.sh
+++ b/test/prepare.sh
@@ -40,13 +40,44 @@ popd
 
 # Installing MariaDB operator
 
-"${HELM}" repo add mariadb-operator https://helm.mariadb.com/mariadb-operator
-"${HELM}" install mariadb-operator-crds mariadb-operator/mariadb-operator-crds \
-    --version "${MARIADB_OPERATOR_VERSION}"
-"${HELM}" install mariadb-operator mariadb-operator/mariadb-operator \
-    --namespace mariadb-operator --create-namespace \
-    --set webhook.cert.certManager.enabled=true \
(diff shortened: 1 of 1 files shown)


test/prepare.sh
```diff
@@ -40,13 +40,44 @@ popd
...
+echo "Installing MariaDB Operator CRDs..."
+for i in {1..5}; do
+    if "${HELM}" install mariadb-operator-crds mariadb-operator/mariadb-operator-crds \
+        --version "${MARIADB_OPERATOR_VERSION}"; then
+        break
+    fi
+    echo "Attempt $i failed, retrying in 10s..."
+    sleep 10
+
+    if [ "$i" -eq 5 ]; then
```
tuminoid:
```suggestion
    if [[ "$i" -eq 5 ]]; then
```

test/prepare.sh
```diff
@@ -40,13 +40,44 @@ popd
...
+        echo "Failed to install MariaDB CRDs after 5 attempts."
```
tuminoid:
```suggestion
        echo "ERROR: Failed to install MariaDB CRDs after 5 attempts."
```

tuminoid (commented):
It is easier for bots etc to detect fatal failures with ERROR.

tuminoid (commented):
/lgtm

PR #553, opened 2026-02-17
✨ Add networkRanges to DHCP for multi-subnet support

holmboe:
## Summary
- Add `DHCPRange` struct and `NetworkRanges []DHCPRange` field to the `DHCP` CRD type
- Operator concatenates primary range + `networkRanges` into a single semicolon-separated `DHCP_RANGE` env var
- Relax validation so primary flat fields (`networkCIDR`/`rangeBegin`/`rangeEnd`) are optional when `networkRanges` is provided
- Each network range is independently validated (CIDR parse, IPs within CIDR)
- Fully backward compatible: existing CRs with only flat fields continue to work

## Companion PR
- metal3-io/ironic-image#918 — template change to split `DHCP_RANGE` on `;` and emit multiple `dhcp-range=` lines

## Test plan
- [x] `make test` passes — includes new validation and `buildDHCPRange` test cases
- [x] Verify CRD schema includes `networkRanges` field
- [x] Apply Ironic CR with `networkRanges` only (no primary range) — validates new code path
- [x] Apply Ironic CR with both primary range and `networkRanges` — verify semicolon-separated `DHCP_RANGE`
- [x] Apply Ironic CR with primary range only — verify backward compatibility


tuminoid (commented):
/ok-to-test

api/v1alpha1/ironic_types.go
dtantsur: How should we handle these fields? Does it make sense to call them deprecated and recommend using `NetworkRanges`?

Rozzii (commented):
@holmboe please re-base the PR

holmboe:
Rebased onto current `main` and added the missing sign-off, so this PR is no longer in `needs-rebase` status.

That said: I want to acknowledge #673 (thanks @ftarasenko) — it's a clear superset of what this PR proposes, with named ranges, per-range gateways, and explicit DHCP relay support. I'm endorsing #673 as the path forward upstream and would suggest reviewers focus their attention there. I'm keeping #553 rebased so I can continue using it in a downstream deployment, but I don't expect it to merge ahead of #673. Happy to close once #673 lands.

Rozzii:
@holmboe please rebase, I will also re-request review from everyone.

dtantsur:
A similar feature was implemented in https://github.com/metal3-io/ironic-standalone-operator/pull/673, should this PR be closed now?

dtantsur:
Superseded by https://github.com/metal3-io/ironic-standalone-operator/pull/673

PR #571, opened 2026-03-04
:seedling: harden github actions workflows

tuminoid:
Add persist-credentials: false to checkout actions, switch pr-verifier from pull_request_target to pull_request, add explicit permissions to workflows, bump actions to latest major versions, suppress accepted findings, and apply zizmor safe fixes.


Rozzii (commented):
/lgtm

PR #598, opened 2026-03-16
:seedling: e2e: download ipa image only once

tuminoid:
Host IPA image locally for upgrade testing to avoid downloading it multiple times which is error prone.

Manual cherry-pick of #575 .


Rozzii (commented):
/approve

PR #600, opened 2026-03-17
🌱 Bump golang to v1.25 in go modules and golangci-lint to v2.10.1

### LINKED ISSUE #580: Bump golang in linting/Prow to go 1.25.x
Just noting that we need to bump linters and prow to 1.25 separately, which requires some fixing. This just bumps the go used for image building.

_Originally posted by @tuminoid in https://github.com/metal3-io/ironic-standalone-operator/pull/576#pullrequestreview-3920321057_

kashifest:
Also fixing linter findings

Fixes #580 

**Checklist:**

### DIFF
--- a/cmd/run-local-ironic/main.go
+++ b/cmd/run-local-ironic/main.go
@@ -106,7 +106,7 @@ func runLocalIronic(inputFile, outputFile string, versionInfo ironic.VersionInfo
 	}
 
 	outputDir := filepath.Dir(outputFile)
-	if err = os.MkdirAll(outputDir, 0o755); err != nil {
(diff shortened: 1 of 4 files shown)


tuminoid:
> I assume we're doing the go.mod bump because it's going to be forced on us by gophercloud? Or are there other reasons?

go 1.24 is EOL and 1.25.0 is forced on us by gophercloud, and in https://github.com/metal3-io/ironic-standalone-operator/pull/602 also by x/crypto etc. The deps will push us to 1.25.0 anyways but at random time, better do it in controlled manner, so the required go stays the same for the lifetime of the release branch (after its cut from main).

PR #619, opened 2026-03-31
⚠️ Require environment label on Secrets and ConfigMaps

lentzi90:
This change is for security. We should not let the high-privilege controller make changes to resources without consent from the resource owner.

User-provided resources must have the environment label set before use. Operator no longer adds the label automatically. Adds tests to verify label enforcement and rejection of unlabeled or incorrectly labeled resources.

This change is to prevent the controller from making changes to user-owned
resources without consent. The label is seen as the consent. For now the controller will still put owner references on these resources, but not if they are missing the label.

It is worth noting also that CAPI does something similar for user-provided certificates: https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates. They must be labeled correctly or the controller will ignore them. 

Fixes #

**Checklist:**

I will push a PR for metal3-docs for this also.

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -11,6 +11,14 @@ const (
 	IronicReasonAvailable  = "DeploymentAvailable"
 
 	IronicLabelPrefix = "ironic.metal3.io"
+
+	// LabelEnvironmentName is the label key that must be present on user-provided
+	// Secrets and ConfigMaps to grant the operator permission to use them.
+	// The value must be LabelEnvironmentValue.
+	LabelEnvironmentName = "environment.metal3.io/ironic-standalone-operator"
+
+	// LabelEnvironmentValue is the required value for the LabelEnvironmentName label.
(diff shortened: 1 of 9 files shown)


tuminoid (commented):
Besides the open comment by @dtantsur , I'm satisfied with the fix. It will make consent explicit and I also agree with the annotation/labels one is enough. It doesn't make any difference in RBACs etc.

/approve
/hold
for merge control.

Rozzii (commented):
/lgtm

lentzi90:
Sorry I forgot to get back to this. Let me move the labels.

dtantsur:
@lentzi90 hmm, shouldn't we update https://github.com/metal3-io/ironic-standalone-operator/blob/main/pkg/ironic/secrets.go#L134-L138 to set the new label on creation? And if yes, how does this change even pass the CI?

dtantsur:
/lgtm

Thanks, unhold when ready.

PR #632, opened 2026-04-08
🌱 Add lint-fix target for auto-fixing linter issues

lentzi90:

Refactor the lint target to support optional arguments via GOLANGCI_LINT_EXTRA_ARGS, enabling a new lint-fix target that runs golangci-lint with the --fix flag across all modules.

Fixes #

**Checklist:**

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,7 @@ Run these locally before submitting PRs:
 | `make generate` | Regenerate DeepCopy methods |
 | `make manifests` | Regenerate CRDs, RBAC, webhooks, API docs |
 | `make lint` | Go linting via golangci-lint (all modules) |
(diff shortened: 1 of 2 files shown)


tuminoid (commented):
/approve

We have this in other controller repos already.

dtantsur:
Sounds nice, but did it actually work in your testing? I've tried a few obvious lint violations, and they all just failed the run.

tuminoid:
> Sounds nice, but did it actually work in your testing? I've tried a few obvious lint violations, and they all just failed the run.

It is bit of a hit and miss, I think it depends on the plugin and its implemented capabilities. Some warnings do clean up automatically, some even obvious ones do not :(

dtantsur:
I see, I suspected this.

/lgtm

PR #635, opened 2026-04-09
✨ Add bindAddress configuration for prometheus exporter

lentzi90:

The first commit just makes the exporter bind to localhost for security. The second commit then makes this configurable, with localhost as default.

This adds a new `bindAddress` field to the PrometheusExporter spec,
allowing operators to configure which IP address the metrics endpoint
listens on. It defaults to "0.0.0.0" for backwards compatibility and
convenience but can be set to a specific IP for added security, or even a loopback
address for specialized security configurations.

Fixes #

**Checklist:**

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -32,7 +32,8 @@ const (
 	certsDir  = "/certs"
 	sharedDir = "/shared"
 
-	metricsPort = 9608
+	metricsPort            = 9608
+	defaultMetricsBindAddr = "0.0.0.0"
 
 	knownExistingPath = "/images/ironic-python-agent.kernel"
 
@@ -667,6 +668,13 @@ func newKeepalivedContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) c
 	}
 }
 
+func flaskRunHost(ironic *metal3api.Ironic) string {
+	if ironic.Spec.PrometheusExporter != nil && ironic.Spec.PrometheusExporter.BindAddress != "" {
+		return ironic.Spec.PrometheusExporter.BindAddress
+	}
+	return defaultMetricsBindAddr
+}
+
 func newPrometheusExporterContainer(versionInfo VersionInfo, ironic *metal3api.Ironic, volumeMount corev1.VolumeMount) corev1.Container {
 	port := int32(metricsPort)
 	if ironic.Spec.Networking.PrometheusExporterPort != 0 {
@@ -679,10 +687,11 @@ func newPrometheusExporterContainer(versionInfo VersionInfo, ironic *metal3api.I
 		Command: []string{"/bin/runironic-exporter"},
 		Env: []corev1.EnvVar{
 			{
-				// Bind to localhost to prevent unauthenticated access from the
-				// network when the pod runs with hostNetwork: true.
+				// Bind address for the metrics endpoint. Defaults to "0.0.0.0" (all interfaces).
+				// Can be overridden via spec.prometheusExporter.bindAddress, e.g. to
+				// "127.0.0.1" to restrict access to the local host only.
 				Name:  "FLASK_RUN_HOST",
-				Value: "127.0.0.1",
+				Value: flaskRunHost(ironic),
 			},
 			{
 				Name:  "FLASK_RUN_PORT",
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -679,6 +679,12 @@ func newPrometheusExporterContainer(versionInfo VersionInfo, ironic *metal3api.I
 		Command: []string{"/bin/runironic-exporter"},
 		Env: []corev1.EnvVar{
 			{
+				// Bind to localhost to prevent unauthenticated access from the
+				// network when the pod runs with hostNetwork: true.
+				Name:  "FLASK_RUN_HOST",
+				Value: "127.0.0.1",
+			},
+			{
 				Name:  "FLASK_RUN_PORT",
 				Value: strconv.Itoa(int(port)),
 			},
--- a/pkg/ironic/servicemonitor.go
+++ b/pkg/ironic/servicemonitor.go
@@ -41,6 +41,10 @@ func ensureServiceMonitor(cctx ControllerContext, ironic *metal3api.Ironic) (Sta
 			{
 				Port: metricsPortName,
 				// TODO(dtantsur): TLS support?
+				// NOTE: The ironic-prometheus-exporter binds to localhost (127.0.0.1) by default
+				// to prevent unauthenticated access from the host network. This means that
+				// ServiceMonitor-based scraping from a remote Prometheus instance will NOT work
+				// unless the exporter is explicitly configured to listen on all interfaces.
 				Scheme: ptr.To(monitoringv1.SchemeHTTP),
 				Path:   "/metrics",
 			},
--- a/pkg/ironic/servicemonitor.go
+++ b/pkg/ironic/servicemonitor.go
@@ -41,10 +41,6 @@ func ensureServiceMonitor(cctx ControllerContext, ironic *metal3api.Ironic) (Sta
 			{
 				Port: metricsPortName,
 				// TODO(dtantsur): TLS support?
-				// NOTE: The ironic-prometheus-exporter binds to localhost (127.0.0.1) by default
-				// to prevent unauthenticated access from the host network. This means that
-				// ServiceMonitor-based scraping from a remote Prometheus instance will NOT work
-				// unless the exporter is explicitly configured to listen on all interfaces.
 				Scheme: ptr.To(monitoringv1.SchemeHTTP),
 				Path:   "/metrics",
 			},
--- a/pkg/ironic/validation.go
+++ b/pkg/ironic/validation.go
@@ -3,6 +3,7 @@ package ironic
 import (
 	"errors"
 	"fmt"
+	"net"
 	"net/netip"
 	"net/url"
 	"reflect"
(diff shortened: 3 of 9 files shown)


lentzi90:
Hmm maybe we should mark it as :warning: because this changes the value from 0.0.0.0 to 127.0.0.1 unless the user takes action?

dtantsur:
@lentzi90 it's breaking for sure... I wonder if the default of 127.0.0.1 even makes any sense or that we should add a big ugly warning to the `enabled` flag instead. In other words, is there a valid case of setting `enabled` to `true` but not keeping the default?

pkg/ironic/containers.go
```diff
@@ -678,6 +686,14 @@ func newPrometheusExporterContainer(versionInfo VersionInfo, ironic *metal3api.I
 		Image:   versionInfo.IronicImage,
 		Command: []string{"/bin/runironic-exporter"},
 		Env: []corev1.EnvVar{
+			{
+				// Bind to localhost by default to prevent unauthenticated access from the
+				// network when the pod runs with hostNetwork: true.
+				// Can be overridden via spec.prometheusExporter.bindAddress, e.g. to
+				// "0.0.0.0" for ServiceMonitor-based scraping from remote Prometheus instances.
+				Name:  "FLASK_RUN_HOST",
+				Value: flaskRunHost(ironic),
```
dtantsur, line 694: Remind me please: can containers get their cluster IP when host networking is used, or will it always be the host IP?

pkg/ironic/servicemonitor.go
```diff
@@ -41,6 +41,10 @@ func ensureServiceMonitor(cctx ControllerContext, ironic *metal3api.Ironic) (Sta
 			{
 				Port: metricsPortName,
 				// TODO(dtantsur): TLS support?
+				// NOTE: The ironic-prometheus-exporter binds to localhost (127.0.0.1) by default
+				// to prevent unauthenticated access from the host network. This means that
+				// ServiceMonitor-based scraping from a remote Prometheus instance will NOT work
+				// unless spec.prometheusExporter.bindAddress is set to a non-loopback address.
```
dtantsur: If it does not work, we should not allow this configuration.

pkg/ironic/containers.go
lentzi90: I think it is always just the host IP unfortunately

lentzi90:
Right, I did not consider that. I guess it would be ok to just make it disabled by default and add some warnings around enabling it. That would be a far simpler change.

> is there a valid case of setting enabled to true but not keeping the default?
 Nope, and I actually added validation in the webhook to avoid any ambiguity around this.

@tuminoid what do you think about just making the exported disabled by default, vs making it listen on localhost?

pkg/ironic/servicemonitor.go
lentzi90: Indeed. I added validation for it but still felt the need to mention this here :upside_down_face:

pkg/ironic/validation.go
```diff
@@ -215,6 +216,24 @@ func ValidateIronic(ironic *metal3api.IronicSpec, old *metal3api.IronicSpec) err
...
 	}
 
+	if ironic.PrometheusExporter != nil && ironic.PrometheusExporter.BindAddress != "" {
+		ip := net.ParseIP(ironic.PrometheusExporter.BindAddress)
+		if ip == nil {
+			return fmt.Errorf("prometheusExporter: bindAddress %q is not a valid IP address", ironic.PrometheusExporter.BindAddress)
+		}
+	}
+
+	if ironic.PrometheusExporter != nil && !ironic.PrometheusExporter.DisableServiceMonitor {
```
tuminoid:
```suggestion
	if ironic.PrometheusExporter != nil && ironic.PrometheusExporter.Enabled && !ironic.PrometheusExporter.DisableServiceMonitor {
```

pkg/ironic/validation_test.go
```diff
@@ -339,11 +339,78 @@ func TestValidateIronic(t *testing.T) {
...
+		{
+			Scenario: "invalid bindAddress",
+			Ironic: metal3api.IronicSpec{
+				PrometheusExporter: &metal3api.PrometheusExporter{
+					Enabled:     true,
+					BindAddress: "not-an-ip",
+				},
+			},
+			ExpectedError: "bindAddress \"not-an-ip\" is not a valid IP address",
+		},
```
tuminoid, line 450: We need test scenario: "disabled exporter with default bindAddress passes validation"

test/suite_test.go
```diff
@@ -466,16 +469,36 @@ func verifyConductorList(ctx context.Context, cli *gophercloud.ServiceClient, as
...
+	if !isLoopback {
+		By(fmt.Sprintf("checking ironic-prometheus-exporter is accessible via node IP (bindAddress=%s)", bindAddress))
 
-	// NOTE(dtantsur): each Ironic replica has its own exporter, so verify them all independently
-	for _, ironicIP := range currentIronicIPs {
-		testURL := fmt.Sprintf("http://%s/metrics", net.JoinHostPort(ironicIP, "9608"))
-		statusCode := helpers.GetStatusCode(ctx, &httpClient, testURL)
-		Expect(statusCode).To(Equal(200))
+		httpClient := helpers.NewHTTPClient()
+
```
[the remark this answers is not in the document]
tuminoid: This is legit, we need test case for specific ip bind.

tuminoid (commented):
> @tuminoid what do you think about just making the exported disabled by default, vs making it listen on localhost?

Disabled by default as we have three legit enabled configuration here, and we need user to make the selection.
 - Localhost scraping → enabled: true, default bind, disableServiceMonitor: true
 - Remote Prometheus → enabled: true, bindAddress: "0.0.0.0", ServiceMonitor on
 - Specific interface → enabled: true, specific IP, disableServiceMonitor: true

This requires user action when they have `enabled: true`, so ⚠️ would be right IMO.

dtantsur:
> Localhost scraping → enabled: true, default bind, disableServiceMonitor: true

What's the user story here? We cannot just assume that prometheus always runs on the same node as Ironic.

I'd rather add a big scary warning to `enabled` (maybe we should have added a feature gate for this thing..)

tuminoid:
> > Localhost scraping → enabled: true, default bind, disableServiceMonitor: true
> 
> What's the user story here? We cannot just assume that prometheus always runs on the same node as Ironic.

Nobody is assuming, the whole point was to force user configure it, if they set enabled: true

edit: ah, I said default bind, whoops. The point is to have user configure the IP, if its 127.0.0.1 or some other IP, so case 1 and 3 are practically the same.

> I'd rather add a big scary warning to `enabled` (maybe we should have added a feature gate for this thing..)

I don't think we need feature gate, since we default to disabled.

lentzi90:
Wait wait, let's take a step back here. We are talking about all different things at once.

The goal with this PR is to get a secure default for the exporter. My first commit was then to make it bind to localhost.
That is a pretty useless configuration though, so then I introduced the feature to make it configurable. What I didn't consider at the time, is that I could have just switched the default on exporter to default off. I am not sure there is a demand for having the bind address configurable at all, it was just my solution to the "issue" that came with how I secured the default config.

In other words, I am very happy to just scrap this feature and simply set the default to disable the exporter, plus a warning for users about what it means to enable it.

Do you think there is reason to keep the bind address configurable?

tuminoid:
> Do you think there is reason to keep the bind address configurable?

If its not configurable, the value would be then servicemonitor compatible 0.0.0.0, bound in host network? I'd think that is not ok either for everybody.

dtantsur:
I'm fine with any reasonable default value, but I'm not fine with deliberately breaking the feature and making users to do a double opt-in.

dtantsur:
> What I didn't consider at the time, is that I could have just switched the default on exporter to default off. 

The exporter is already disabled by default, so I'm somewhat confused by this comment.

lentzi90:
> > What I didn't consider at the time, is that I could have just switched the default on exporter to default off.
> 
> The exporter is already disabled by default, so I'm somewhat confused by this comment.

Oh, sorry. I confused myself. Never mind this.

So the question is then that what would be a "reasonable" default bind address? Localhost without host network is unusable, but since we have host network I guess it is at least in theory useful.

Would there be any point binding to some address that Ironic uses?

pkg/ironic/validation.go
```diff
@@ -215,6 +216,24 @@ func ValidateIronic(ironic *metal3api.IronicSpec, old *metal3api.IronicSpec) err
...
+		bindAddr := ironic.PrometheusExporter.BindAddress
+		if bindAddr == "" {
+			bindAddr = defaultMetricsBindAddr
+		}
```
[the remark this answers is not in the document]
lentzi90: Fixed

test/suite_test.go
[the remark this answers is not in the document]
lentzi90: Fixed

pkg/ironic/validation.go
lentzi90: Done

pkg/ironic/validation_test.go
lentzi90: Added

lentzi90:
/retitle ⚠️ Default prometheus exporter bindAddress to loopback and make it configurable

dtantsur (changes requested):
I'm fine with bindAddress being configurable, but I'm against the default that won't work for most applications. The exporter support is already opt-in. If people ignore one warning about security implications, they'll blindly set as many values as you need.

Imagine our user guide. Instead of showing:

```yaml
spec:
  prometheusExporter:
    enabled: true
```

it will show

```yaml
spec:
  prometheusExporter:
    enabled: true
    bindAddress: 0.0.0.0
```

So, what have we achieved? Users will copy-paste it into their deployment anyway. (And no, I'm not going to skip bindAddress from the docs: we don't cripple our documentation on purpose.)

tuminoid:
> So, what have we achieved? Users will copy-paste it into their deployment anyway. (And no, I'm not going to skip bindAddress from the docs: we don't cripple our documentation on purpose.)

I can live with this, even if I don't love it.

lentzi90:
Ok, so the conclusion is to make it configurable but keep the wildcard default.
I'll update. This also means then that it is no longer a :warning: but instead :sparkles:

lentzi90:
/retitle ✨ Add bindAddress configuration for prometheus exporter

test/suite_test.go
```diff
@@ -1183,10 +1225,72 @@ var _ = Describe("Ironic object tests", func() {
...
+		DeferCleanup(func() {
+			CollectLogs(namespace)
+			DeleteAndWait(ironic)
+		})
+
+		ironic = WaitForIronic(name)
+		VerifyIronic(ironic, TestAssumptions{withPrometheusExporter: true, exporterBindAddress: "0.0.0.0"})
+	})
+
+	It("creates Ironic with prometheus exporter bound to wildcard address", Label("prometheus-exporter-wildcard"), func() {
```
dtantsur: This test seems identical to the previous one now

dtantsur (dismissed):

test/suite_test.go
lentzi90: True! Fixed now

tuminoid (commented):
/approve

PR #636, opened 2026-04-13
✨ Add Keepalived list and fields

### LINKED ISSUE #490: Support separate network for HTTP service
It is quite common that a separate out of band (OOB) network is used for BMCs. This can make it tricky to use virtualmedia with Ironic when deployed by IrSO. I want to support the following configuration:

- Provisioning network: 192.168.0.0/24 - no VLAN. Provisioning IP belongs here, dnsmasq provides DHCP here.
- OOB network: 192.168.1.0/24 - VLAN 2. No DHCP. BMCs live here. Ironic needs access to this network to communicate with them. IRONIC_HTTP_URL belongs here since the BMCs need to access virtualmedia through it.

Before IrSO, I have been using IRONIC_HTTP_URL to achieve this. I am not sure if it is the "correct" way, but it works. I simply make sure that the node where Ironic runs has an IP assigned that I can use and set the variable to it. Our keepalived image [only supports a single IP](https://github.com/metal3-io/utility-images/blob/0f6fce5c0156e6ca199d930424dccbb72b9b0922/keepalived/manage-keepalived.sh#L15) currently so I have been doing it in other ways. However, I am hoping that we could add this also as a feature there.

I do not have a complete design ready in my head, but here is something as a discussion starter:

```yaml
apiVersion: ironic.metal3.io/v1alpha1
kind: Ironic
metadata:
  name: ironic
  namespace: baremetal-operator-system
spec:
  networking:
    dhcp:
      rangeBegin: "192.168.0.200"
      rangeEnd: "192.168.0.220"
      networkCIDR: "192.168.0.0/24"
    interface: "eth0"
    ipAddress: "192.168.0.100"
    ipAddressManager: "keepalived"
    imageServerIpAddress: "192.168.1.50" # new field
    # imageServerPort: 80 # already implemented
    # imageServerTLSPort: 443 # already implemented
    keepalived: # new struct, requires support in keepalived scripts
      - ipAddress: 192.168.0.100
        interface: "eth0"
      - ipAddress: 192.168.1.50
        interface: "eth1"
    # keepalivedOverrideConfig: | # alternative escape hatch
      # vrrp_instance V_1 {
      #   state MASTER
      # ...
```

lentzi90:

Add support for managing multiple virtual IPs via a new `keepalived`
field as an alternative to the existing `ipAddressManager` and deprecate
the latter. This enables
more advanced Keepalived configurations.

When Keepalived is enabled, it will use the interface and IP address
from the `networking` field. It will also take the prefix from the DHCP
network CIDR if specified. Additional addresses, interfaces and prefixes
can be specified under the `keepalived` field.

NOTE: This requires keepalived image sha256:07835bae6e8ca62f83292da94f2bf76fe20182e18516deb89c109bc7b02ac1b0 (built 2026-03-02) or newer.
The functionality builds on the feature in https://github.com/metal3-io/utility-images/pull/51.

Fixes #490

**Checklist:**

- [x] Documentation has been updated, if necessary. (The auto-generated API docs that is.)


pkg/ironic/validation_test.go
```diff
@@ -199,6 +199,144 @@ func TestValidateIronic(t *testing.T) {
...
+		},
+		{
+			Scenario: "Keepalived list and ipAddressManager are mutually exclusive",
+			Ironic: metal3api.IronicSpec{
+				Networking: metal3api.Networking{
+					Interface:        "eth0",
+					IPAddress:        "192.0.2.2",
+					IPAddressManager: metal3api.IPAddressManagerKeepalived,
+					Keepalived: []metal3api.KeepalivedIP{
+						{IPAddress: "192.0.2.2", Interface: "eth0"},
```
lentzi90: This could perhaps be confusing. Basically the user has to choose between the old, simple toggle and the new struct with more options. To keep things simple, you cannot mix these.

pkg/ironic/validation_test.go
```diff
@@ -199,6 +199,144 @@ func TestValidateIronic(t *testing.T) {
...
+			Scenario: "Keepalived list requires interface",
+			Ironic: metal3api.IronicSpec{
+				Networking: metal3api.Networking{
+					IPAddress: "192.0.2.2",
+					Keepalived: []metal3api.KeepalivedIP{
+						{IPAddress: "192.0.2.2", Interface: "eth0"},
+					},
+				},
+			},
+			ExpectedError: "keepalived requires specifying both ipAddress and interface",
```
lentzi90: I want to call out these two tests explicitly. I am not too happy with the behavior here but I am not sure how to do it better.

The issue is that, when using keepalived, we need to know the IP/interface for both Ironic and Keepalived. This ends up as duplicate information, which could be annoying and confusing for users.

The old way of just enabling keepalived, and then use the IP/interface of Ironic is nicer in this way. We could default to include Ironic's IP/interface also for keepalived, but then how do we tell when it is enabled or not? I guess we would still need ipAddressManager for that. Plus make it obvious to users that there will be some defaulting happening.

lentzi90:
This PR currently implements something like this:

```yaml
apiVersion: ironic.metal3.io/v1alpha1
kind: Ironic
metadata:
  name: ironic
  namespace: baremetal-operator-system
spec:
  networking:
    dhcp:
      rangeBegin: "192.168.0.200"
      rangeEnd: "192.168.0.220"
      networkCIDR: "192.168.0.0/24"
    interface: "eth0"
    ipAddress: "192.168.0.100"
    # ipAddressManager: "keepalived" # the old way of enabling keepalived
    imageServerIpAddress: "192.168.1.50" # new field, name following existing conventions
    # imageServerPort: 80 # already existing field
    # imageServerTLSPort: 443 # already existing field
    keepalived: # new struct
      - ipAddress: 192.168.0.100
        interface: "eth0"
      - ipAddress: 192.168.1.50
        interface: "eth1"
```

After implementing the above, I started thinking about alternatives. I put some comments above to call out things I wanted to improve. Based on that, I came up with the following. What do you think about this? Is it better?

```yaml
apiVersion: ironic.metal3.io/v1alpha1
kind: Ironic
metadata:
  name: ironic
  namespace: baremetal-operator-system
spec:
  networking:
    dhcp:
      rangeBegin: "192.168.0.200"
      rangeEnd: "192.168.0.220"
      networkCIDR: "192.168.0.0/24"
    interface: "eth0"
    ipAddress: "192.168.0.100"
    imageServerIpAddress: "192.168.1.50" # New field
    keepalived: # New struct
      enabled: true # Explicitly enabled to avoid need for ipAddressManager
      additionalVIPs: # Only list additional VIPs to avoid duplicating spec.networking
      # This is always included, taken from spec.neworking
      # - ipAddress: 192.168.0.100
      #   interface: "eth0"
      - ipAddress: 192.168.1.50
        interface: "eth1"
```

api/v1alpha1/ironic_types.go
```diff
@@ -151,11 +169,20 @@ type Networking struct {
...
+	// Also see the keepalived field for an alternative with more advanced options.
 	// Warning: keepalived is not compatible with the highly available architecture.
 	// +kubebuilder:validation:Enum="";keepalived
 	// +optional
 	IPAddressManager IPAddressManager `json:"ipAddressManager,omitempty"`
 
+	// Keepalived is a list of virtual IPs to be managed by Keepalived.
+	// When set, a Keepalived container will be started to manage these IPs.
+	// The first entry's interface is used as the primary VRRP interface.
+	// Cannot be used together with ipAddressManager.
```
dtantsur, line 196: Given that no other IP address managers are in sight, I think we need to deprecate `ipAddressManager` entirely and rely on a new substructure like

```yaml
spec:
  networking:
    keepalived:
      enabled: true
      interfaces: []  # optional
```

Thoughts?

pkg/ironic/containers.go
```diff
@@ -639,15 +665,28 @@ func newDnsmasqContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) core
...
+
+	if len(ironic.Spec.Networking.Keepalived) > 0 {
+		// New multi-IP mode: use KEEPALIVED_VIRTUAL_IPS env var.
+		// Format: space-separated "ip,interface" entries.
+		entries := make([]string, 0, len(ironic.Spec.Networking.Keepalived))
+		for _, vip := range ironic.Spec.Networking.Keepalived {
+			entries = append(entries, fmt.Sprintf("%s,%s", vip.IPAddress, vip.Interface))
+		}
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "KEEPALIVED_VIRTUAL_IPS",
```
dtantsur, line 744: We don't version the keepalived container, right? So this will work with any versions?

pkg/ironic/validation_test.go
dtantsur: Yeah, see above about a potential deprecation

api/v1alpha1/ironic_types.go
lentzi90: Yeah, I can agree to this. One question though: do we make the user list all interfaces/IPs under keepalived, or do we just require additional interfaces/IPs to those that are already set in `spec.networking`?
I am thinking users might get annoyed if they have to put them in two places just to enable keepalived.

pkg/ironic/containers.go
lentzi90: That is a good point. It is true that we don't version it. It will work with any image built after https://github.com/metal3-io/utility-images/pull/51 was merged.
Any suggestions for how to handle this?

api/v1alpha1/ironic_types.go
dtantsur: Hmm. Add by default but add `skipDefaultInterface` if they need to override everything? Is there a valid use case for not including the default interface or am I overthinking it?

pkg/ironic/containers.go
dtantsur: We default to the latest tag, don't we? Which mean it will be reloaded on the next update?

dtantsur: For the future, I wonder if we need to bind keepalived to a specific hash.

lentzi90: Yeah we default to latest in https://github.com/metal3-io/ironic-standalone-operator/blob/2658639354da6a33448cc01a652b6e0ce65f171a/pkg/ironic/version.go#L19.
Pinning this (and all other) per Ironic release sounds very reasonable.
I'll check the hash that is needed and add to the PR description so users can pin if wanted.

api/v1alpha1/ironic_types.go
lentzi90: I cannot think of a reason to not include. I'll try it out and see how it feels. Putting this PR as WIP in the mean time.

lentzi90:
/hold
Trying some different options.

pkg/ironic/validation_test.go
```diff
@@ -199,6 +199,144 @@ func TestValidateIronic(t *testing.T) {
...
+				HighAvailability: true,
+				Networking: metal3api.Networking{
+					Interface: "eth0",
+					IPAddress: "192.0.2.2",
+					Keepalived: []metal3api.KeepalivedIP{
+						{IPAddress: "192.0.2.2", Interface: "eth0"},
+					},
+				},
+			},
+			ExpectedError: "ipAddress makes no sense with highly available architecture",
```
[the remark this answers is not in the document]
lentzi90: Addressed by adding a "no ipAddress" test case that reaches the new branch

lentzi90:
/hold cancel
I am quite happy with this now.

pkg/ironic/containers.go
```diff
@@ -364,6 +369,27 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
 	return result
 }
 
+func buildImageServerURLEnvVar(ironic *metal3api.Ironic) []corev1.EnvVar {
```
tuminoid: We return array of envvars for single element?

tuminoid (commented):
A question, looks clean otherwise.

pkg/ironic/containers.go
lentzi90: That's a good point. I think all the other functions return multiple so I didn't think much about it. Changing to single EnvVar

lentzi90: Fixed

api/v1alpha1/ironic_types.go
```diff
@@ -132,6 +164,13 @@ type Networking struct {
 	// +optional
 	ImageServerPort int32 `json:"imageServerPort,omitempty"`
 
+	// ImageServerIPAddress is the IP address from which BMCs will access the image server
+	// for virtual media. Use this when BMCs live on a separate network (e.g., out-of-band management)
+	// and need to access virtual media images through a different IP than the main provisioning IP.
+	// When not set, the main IPAddress (or ExternalIP if set) is used.
+	// +optional
+	ImageServerIPAddress string `json:"imageServerIPAddress,omitempty"`
```
dtantsur: This change conflicts with https://github.com/metal3-io/ironic-standalone-operator/pull/655, so I wonder if you can omit it for now and focus on keepalived only

1nv0k32: Moved those changes to #679

lentzi90: I can take it out of this PR regardless. It can be done separately :slightly_smiling_face:

kashifest (approved):
lgtm

tuminoid (commented):
/approve

Clean to me as well.

PR #644, opened 2026-04-23
:sparkles: Support ironic-image 35.0

hroyrh:
Add support for ironic-image 35.0.

### DIFF
--- a/test/suite_test.go
+++ b/test/suite_test.go
@@ -65,8 +65,9 @@ const (
 	apiVersionIn320 = "1.101"
 	apiVersionIn330 = "1.104"
 	apiVersionIn340 = "1.109"
+	apiVersionIn350 = "1.111"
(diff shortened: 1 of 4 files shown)


test/suite_test.go
```diff
@@ -65,8 +65,9 @@ const (
 	apiVersionIn320 = "1.101"
 	apiVersionIn330 = "1.104"
 	apiVersionIn340 = "1.109"
+	apiVersionIn350 = "1.111"
 	// Update this periodically to make sure we're installing the latest version by default.
-	knownAPIMinorVersion = 109
+	knownAPIMinorVersion = 111
```
dtantsur: Latest Ironic is 1.112 already. Which is handy, because we can distinguish between 35 and latest in this test.

hroyrh: Makes sense, updated it.

Rozzii (commented):
/approve

PR #646, opened 2026-04-27
:sparkles: Switch to ironic 35.0 by default, prepare release-0.9

hroyrh:


test/suite_test.go
```diff
@@ -66,7 +66,8 @@ const (
 	apiVersionIn340 = "1.109"
 	apiVersionIn350 = "1.111"
 	// Update this periodically to make sure we're installing the latest version by default.
-	knownAPIMinorVersion = 112
+	// But in case of a release, keep this same as the release supported api version.
```
hroyrh, line 69: Added a comment here that we don't use the latest supported API version here.

Rozzii (commented):
/approve

PR #651, opened 2026-04-30
🌱 Fix IRSO functional test 

smoshiur1237:
The test is checking if the prometheus exporter is accessible, but it's not using a retry mechanism (Eventually) to handle the case where the service might not be immediately ready to accept connections. So this PR will handle the situation by adding eventually to add retry option.

### DIFF
--- a/test/helpers/clients.go
+++ b/test/helpers/clients.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -48,10 +49,18 @@ func NewHTTPClient() http.Client {
 
 func GetStatusCode(ctx context.Context, httpClient *http.Client, url string) int {
 	req, err := http.NewRequestWithContext(ctx, http.MethodHead, url, http.NoBody)
-	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
+		GinkgoWriter.Printf("Failed to create request for %s: %v\n", url, err)
+		// Return 0 to allow Eventually to retry on errors
+		return 0
+	}
 
 	resp, err := httpClient.Do(req) //nolint:gosec // URL is controlled by test infrastructure
-	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
+		GinkgoWriter.Printf("Failed to connect to %s: %v\n", url, err)
+		// Return 0 to allow Eventually to retry on connection errors
+		return 0
+	}
 	defer resp.Body.Close()
 
 	return resp.StatusCode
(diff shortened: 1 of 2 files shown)


smoshiur1237:
/cc @tuminoid @lentzi90 could you check if it help with the IRSO functional test in ironic-image.
https://github.com/metal3-io/ironic-image/actions/runs/25147623780/job/73720552580?pr=981

dtantsur:
@smoshiur1237 hmm, I can accept this as a *temporary* workaround, but I think the real fix should be having a readiness probe for IPE. Since IPE supports neither authentication nor TLS at the moment, you could add a simple HTTP probe.

smoshiur1237:
> @smoshiur1237 hmm, I can accept this as a _temporary_ workaround, but I think the real fix should be having a readiness probe for IPE. Since IPE supports neither authentication nor TLS at the moment, you could add a simple HTTP probe.

As all the ironic-image PR test for IRSO are failing, if you can take the workaround. I will try to add the rediness probe for IPE in separate PR latest next week.

dtantsur:
Do we at least have understanding why it started failing all of a sudden? The prometheus code is not new.

smoshiur1237:
This is a test robustness fix, not a bug in the actual prometheus exporter functionality. The exporter works correctly; the test just needed to account for natural startup timing variations.

Here is the git commits: 
```
# The feature was added without Eventually:
fc0edf3 - Add bindAddress configuration for prometheus exporter (Apr 17)

# Tests for ironic-image 35.0 were added:
400e2cf - Support ironic-image 35.0 (Apr 23)

# fix (HEAD):
7d21c8a - Fix functional test (Apr 30) ← You're fixing this exact issue!

```
I think this is a classic race condition that was always present but only surfaced intermittently:

```
Pod Ready ≠ Service Ready: The Kubernetes pod reaches "Ready" status before the prometheus exporter Flask application is fully accepting connections on port 9608

Timing Window: There's a small time window (usually 1-5 seconds) between:

The Ironic deployment being marked as Ready
The prometheus exporter actually binding to the network interface and accepting connections
CI Environment Variability: In CI environments with varying loads:

Sometimes the exporter starts fast enough → test passes ✅
Sometimes there's a slight delay → connection refused ❌
```

test/helpers/clients.go
```diff
@@ -48,10 +48,16 @@ func NewHTTPClient() http.Client {
 
 func GetStatusCode(ctx context.Context, httpClient *http.Client, url string) int {
 	req, err := http.NewRequestWithContext(ctx, http.MethodHead, url, http.NoBody)
-	Expect(err).NotTo(HaveOccurred())
+	if err != nil {
+		// Return 0 to allow Eventually to retry on errors
```
dtantsur, line 54: Let's at least log errors, otherwise it will be impossible to debug.

dtantsur:
> The feature was added without Eventually:
> fc0edf3 - Add bindAddress configuration for prometheus exporter (Apr 17)

This is not when the feature was added though, it just added bindAddress.

dtantsur:
/lgtm

But please don't forget to follow up (or create a bug for someone else to follow up)

tuminoid (commented):
/approve

Let's get the ironic-image testable again.

smoshiur1237:
> /lgtm
> 
> But please don't forget to follow up (or create a bug for someone else to follow up)

I will push the rediness probe check that you have suggested

smoshiur1237:
IRSO functional test is passing now: https://github.com/metal3-io/ironic-image/actions/runs/25295854668/job/74228708472?pr=984

@tuminoid @dtantsur

PR #655, opened 2026-05-04
✨ Add ingress resource for Ironic, add image server service resource

1nv0k32:

Currently IrSO, doesn't create service for image server.
Also, creating Ingress service is required if it's needed to deploy Metal3 on a cluster which is not directly connected to OOB network (For virtual media deployments)

**Checklist:**


api/v1alpha1/ironic_types.go
```diff
@@ -97,6 +97,26 @@ type DHCP struct {
 	ServeDNS bool `json:"serveDNS,omitempty"`
 }
 
+// Ingress defines ingress resource for Ironic services
+type Ingress struct {
+	// IngressClass of Ingress resource
+	IngressClassName string `json:"ingressClassName,omitempty"`
+
+	// Annotations to be added to Ingress resource
```
dtantsur, line 102: nit: add `+optional` for all optional fields

api/v1alpha1/ironic_types.go
```diff
@@ -97,6 +97,26 @@ type DHCP struct {
...
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Host of the Ingress
```
dtantsur: All docstrings need to be expanded with more details: think less about "what" and more about "why" (as an operator, why do I change this value?)

api/v1alpha1/ironic_types.go
```diff
@@ -97,6 +97,26 @@ type DHCP struct {
...
+	Host string `json:"host,omitempty"`
+
+	// Ironic API Path
+	// +kubebuilder:default=/
```
dtantsur: Does it actually make sense to change this? Will it even work with the current ironic-image?

api/v1alpha1/ironic_types.go
```diff
@@ -97,6 +97,26 @@ type DHCP struct {
...
+// Ingress defines ingress resource for Ironic services
+type Ingress struct {
+	// IngressClass of Ingress resource
+	IngressClassName string `json:"ingressClassName,omitempty"`
+
+	// Annotations to be added to Ingress resource
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Host of the Ingress
+	Host string `json:"host,omitempty"`
```
dtantsur: It's an IP in the current ironic-image design, not a host name, right?

pkg/ironic/ironic.go
```diff
@@ -132,6 +135,12 @@ func ensureIronicService(cctx ControllerContext, ironic *metal3api.Ironic) (Stat
 				Port:       exposedPort,
 				TargetPort: intstr.FromString(ironicPortName),
 			},
+			{
+				Name:       imagesPortName,
+				Protocol:   corev1.ProtocolTCP,
+				Port:       defaultImageExposedPort,
```
dtantsur: This is not going to work correctly in the HA case: Kubernetes expects all backends to be equivalent, which is not the case for Ironic (every image server hosts images only for its instance). I'm not sure how to go around this problem.

api/v1alpha1/ironic_types.go
1nv0k32: Yes, the `ExternalIP`.
In our deployment, we overridden `external_callback_url` and `external_http_url` to the address of the ingress.
If we introduce the ingress, by enabling it, we should propagate the `Host` field instead of IP.

pkg/ironic/ironic.go
1nv0k32: For now dodging it by checking if HA is enabled :D

api/v1alpha1/ironic_types.go
1nv0k32: Just wanted to give the flexibility in case it's needed to change the ingress paths.
But as far as I know, it wouldn't work, specially for `/` as Ironic is using root for version info and if changed, agent will stop working :D

api/v1alpha1/ironic_types.go
1nv0k32: Now I'm configuring ironic-image vars in case ingress is enabled.
Based on this script:
https://github.com/metal3-io/ironic-image/blob/main/scripts/configure-ironic.sh#L57

api/v1alpha1/ironic_types.go
1nv0k32: It didn't make sense to have them, removed.

api/v1alpha1/ironic_types.go
1nv0k32: Made them mode descriptive :D

api/v1alpha1/ironic_types.go
```diff
@@ -121,11 +136,23 @@ type Networking struct {
...
 	DHCP *DHCP `json:"dhcp,omitempty"`
 
+	// Configure Ingress resource for Ironic services.
+	// Set this option when you are planning to deploy Ironic in a public cluster and willing to use Ingress instead of IP address and NodePort.
+	// +optional
+	Ingress *Ingress `json:"ingress,omitempty"`
+
 	// ExternalIP is used for accessing API and the image server from remote hosts.
 	// This settings only applies to virtual media deployments. The IP will not be accessed from the cluster itself.
+	// In case of settings up Ingress, this would be ignored.
```
dtantsur: Then let's add a validation that both are not set together

api/v1alpha1/ironic_types.go
```diff
@@ -121,11 +136,23 @@ type Networking struct {
...
 	// +optional
 	ExternalIP string `json:"externalIP,omitempty"`
 
+	// External HTTP URL for Image server.
+	// Set this option when you're image server is not directly accessible.
+	// Setting this option, will override URL set by Networking.Ingress.Host.
+	// +optional
+	ImageServerExternalURL string `json:"imageServerExternalURL,omitempty"`
```
dtantsur: Not sure how I feel about this one. It interferes with ExternalIP, and we need to decide for a single approach. Do you actually need this option or adding it as a courtesy?

api/v1alpha1/ironic_types.go
```diff
@@ -97,6 +97,21 @@ type DHCP struct {
...
+type Ingress struct {
+	// Annotations to be added to Ingress resource
+	// +optional
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// IngressClass of Ingress resource
+	// +optional
+	IngressClassName string `json:"ingressClassName,omitempty"`
+
+	// The host which Ingress would apply on.
```
dtantsur: This is confusing, could you elaborate? Is it something we can auto-detect instead of making a user provide it?

pkg/ironic/containers.go
```diff
@@ -345,6 +345,16 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
 
 	result = appendStringEnv(result, "IRONIC_EXTERNAL_IP", resources.Ironic.Spec.Networking.ExternalIP)
 
+	if resources.Ironic.Spec.Networking.Ingress != nil {
+		ingressHost := resources.Ironic.Spec.Networking.Ingress.Host
+		result = appendStringEnv(result, "IRONIC_EXTERNAL_CALLBACK_URL", ingressHost)
+		result = appendStringEnv(result, "IRONIC_EXTERNAL_HTTP_URL", ingressHost)
+	}
+
+	if resources.Ironic.Spec.Networking.ImageServerExternalURL != "" {
```
dtantsur: If you decide to keep this option, add a validation that it's not set together with Ingress since they conflict

pkg/ironic/containers.go
```diff
@@ -345,6 +345,16 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
...
+		result = appendStringEnv(result, "IRONIC_EXTERNAL_HTTP_URL", resources.Ironic.Spec.Networking.ImageServerExternalURL)
```
dtantsur: Setting IRONIC_EXTERNAL_HTTP_URL without IRONIC_EXTERNAL_CALLBACK_URL can yield surprising results. See above: maybe we need to generalize ExternaIP to hosts or even host:port pairs?

api/v1alpha1/ironic_types.go
1nv0k32: Oops, the doc is not complete :D
But regarding auto-detecting, not sure if I understand correctly what you mean 🤔 Cause this is the hostname which Ingress will accept requests on, and should be provided by the operator...

api/v1alpha1/ironic_types.go
1nv0k32: We can add checks for either of ExternalIP or Ingress/ImageServerExternalURL.
Regarding `ImageServerExternalURL`, we need it as we cache images per DC on a proxy machine to save bandwidth :D

pkg/ironic/containers.go
1nv0k32: I would say, they can be set together, but setting `ImageServerExternalURL` should override the Ingress.Host as this should be the address of proxy in the OOB network which proxies the calls to Ingress.Host

pkg/ironic/containers.go
1nv0k32: In this case, I'm filling both of them together and shouldn't cause any surprise. In case of generalization, IMO, we should kill ExternalIP in favor of URL (as URLs can be IPs too). Maybe we should deprecate ExternalIP? :thinking:

api/v1alpha1/ironic_types.go
dtantsur: Can we discuss it separately though? We have https://github.com/metal3-io/ironic-standalone-operator/pull/636 from @lentzi90 that introduces more or less the same variable in a slightly different way, and I'd like us to get on the same page without necessarily blocking the rest of Ingress and Keepalived efforts.

1nv0k32: @dtantsur Cleanup anything other than Ingress and image-server service.
I'll create another PR for that.

pkg/ironic/containers.go
1nv0k32: Moved to another PR.

pkg/ironic/containers.go
1nv0k32: Moved to another PR.

dtantsur:
/cc @lentzi90 @kashifest 

... since you probably understand Ingress better than me

pkg/ironic/validation_test.go
```diff
@@ -86,6 +86,18 @@ func TestValidateIronic(t *testing.T) {
 			},
 			ExpectedError: "banana is not a valid IP address",
 		},
+		{
+			Scenario: "ingress and externalIP configured simultaneously",
```
lentzi90, line 100: Maybe add a positive test with just ingress also?

api/v1alpha1/ironic_types.go
```diff
@@ -123,9 +139,16 @@ type Networking struct {
...
 	// This settings only applies to virtual media deployments. The IP will not be accessed from the cluster itself.
+	// Cannot be set at the same time with networking.ingress.
 	// +optional
 	ExternalIP string `json:"externalIP,omitempty"`
 
+	// Configure Ingress resource for Ironic services.
+	// Set this option when you are planning to deploy Ironic in a public cluster and willing to use Ingress instead of IP address and NodePort.
+	// Cannot be set at the same time with networking.externalIP.
+	// +optional
+	Ingress *Ingress `json:"ingress,omitempty"`
```
lentzi90, line 182: I could easily see us add more options in the future. Is there a need to group them? For example, top level field `expose` with sub field `type` that would an enum. Then we could extend from `type: Ingress` to `type: Service` and `type: GatewayAPI` or similar in the future.

From my side, I don't have a strong opinion at this point. We can also just add more top level fields. At some point we can start thinking about a new API version where we clean things up if they get messy.

lentzi90 (commented):
Looks ok to me, but I see tests are failing. Is this ready?
One question that popped up for me: Do you expect to still run Ironic in the host network with this or is that another option you expect to add?

lentzi90 (commented):
/lgtm
Still wondering if we should make it possible to disable host network for these situations? I don't think it makes much sense to use the host network when we expose it through an ingress. In any case, that can be added later.

PR #665, opened 2026-05-12
⚠️ Require environment label on Secrets

lentzi90:
Manual backport of #619
The main difference here is that 0.7 does not deal with configmaps. That was added later.
To keep the changes minimal, I have adjusted the patch to only deal with secrets.

User-provided resources must have the environment label set before use. Operator no longer adds the label automatically. Adds tests to verify label enforcement and rejection of unlabeled or incorrectly labeled resources.

This change is to prevent the controller from making changes to user-owned
resources without consent. The label is seen as the consent. For now the controller will still put owner references on these resources, but not if they are missing the label.

Fixes #

**Checklist:**

### DIFF
--- a/api/v1alpha1/common.go
+++ b/api/v1alpha1/common.go
@@ -11,6 +11,14 @@ const (
 	IronicReasonAvailable  = "DeploymentAvailable"
 
 	IronicLabelPrefix = "ironic.metal3.io"
+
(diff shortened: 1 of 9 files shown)


tuminoid (commented):
/approve
/lgtm

PR #673, opened 2026-05-13
✨ Add multi-subnet DHCP support with per-range options

ftarasenko:

  Add a DHCPRange struct and an extraRanges field to the DHCP CRD type to support serving multiple DHCP subnets from a single dnsmasq instance, typically for subnets reached via DHCP relay agents. Each range specifies its own network CIDR, IP range, and optional gateway address.

  **Key changes**:
  - New DHCPRange type with networkCIDR, rangeBegin, rangeEnd and (optional) gatewayAddress fields. There is no name/tag field — each range is assigned an auto-generated dnsmasq tag range_N (1-based position in the list).
  - Support combining the primary (flat) DHCP range with additional ranges — all are concatenated into the DHCP_RANGE env var as ;-separated entries (extra ranges use set:range_N,...), switching to dotted-netmask notation for IPv4 when extra ranges are present. Single-range output is unchanged.
  - Per-range gateway options via dnsmasq tags, passed through the new DHCP_OPTIONS env var (;-separated). Extra ranges without their own gateway get a valueless tagged option:router to suppress the (unreachable) main-subnet router; IPv6 ranges need no suppression.
  - The readiness/liveness probe checks DHCPv4 (67) and/or DHCPv6 (547) based on the address families across all configured ranges, plus TFTP (69).
  - Primary flat fields (networkCIDR/rangeBegin/rangeEnd) become optional when extraRanges is set (leave all three unset for a relay-only deployment); they are all-or-nothing and remain required when extraRanges is empty. 
  - Validation: per-range CIDR/IP containment, reversed ranges (rangeBegin after rangeEnd), /0 CIDRs, overlapping DHCP pools (main or extra), and rejection of IPv6 per-range gateways. The provisioning IP must still be inside the main CIDR when a main range is configured; extra ranges are not required to contain it (enabling DHCP relay scenarios).
  - Requires Ironic 37.0 or newer (the release adding split DHCP_RANGE/DHCP_OPTIONS support); older versions are rejected.
  - Fully backward compatible: existing CRs with only flat fields work unchanged.

  Supersedes #553

  **Companion PR** 
  - metal3-io/ironic-image#998 — adds `DHCP_OPTIONS` env var support in dnsmasq config template

  **Checklist:**

  - [x] E2E tests have been added, if necessary.


lentzi90 (commented):
/ok-to-test
I need to think how we can fit this together with https://github.com/metal3-io/ironic-standalone-operator/pull/636. If we support multiple DHCP subnets, then it would make sense to have matching support for keepalived.

api/v1alpha1/ironic_types.go
```diff
@@ -91,6 +113,15 @@ type DHCP struct {
...
 	RangeEnd string `json:"rangeEnd,omitempty"`
 
+	// Ranges is a list of DHCP address ranges for multi-subnet support.
+	// Can be used together with the top-level NetworkCIDR, RangeBegin, RangeEnd
+	// fields — both are concatenated into the dnsmasq configuration.
+	// Each range can have its own CIDR and dnsmasq tag name.
+	// The provisioning IP (networking.ipAddress) does not need to be in any range's
+	// CIDR, enabling DHCP relay scenarios.
+	// +optional
+	Ranges []DHCPRange `json:"ranges,omitempty"`
```
dtantsur: If we don't intend to deprecate the existing options, maybe this should be called ExtraRanges or something like that.

That being said, maybe we do need to deprecate these options in favour on `Ranges` with one item? Thoughts? cc @lentzi90

ftarasenko: I'd propose soft-deprecation now, removal later. On rethink, this is cleaner than maintaining two APIs — but nothing breaks on upgrade today.

  - This PR: add // Deprecated: use ranges instead to the godoc on networkCIDR/rangeBegin/rangeEnd/gatewayAddress, and <br>**Deprecated**: use ranges instead. to the CRD description. Existing CRs continue to
  work — ValidateDHCP still accepts the flat path, runtime still builds the same dnsmasq config. Only signal change.
  - Future v1beta1/v1: drop the flat fields. By then users have had time to migrate, and a conversion webhook can rewrite old objects into ranges: [{...}] on read.

ftarasenko:
Infra failed. I'll retry later.
artifactory.nordix.org (129.192.81.245:443) — connection timeout after 75 seconds.

dtantsur:
Once you know in which ironic-image release https://github.com/metal3-io/ironic-image/pull/998 merges, please add version-specific logic. Make sure that you don't try to set DHCP_OPTIONS for versions where it did not exist. You'll probably need to reject multiple ranges on earlier versions.

holmboe:
Thanks for picking up this work, @ftarasenko — and for the explicit "supersedes #553." I've now rebased #553 (mostly for my own downstream deployment use) and want to leave some substantive notes on this PR's design.

### 1. The wire-format design predates ironic-image#918

The code comment in `buildDHCPRange` says:

> The ironic image's dnsmasq.conf.j2 renders: `dhcp-range={{ env.DHCP_RANGE }}`
> so the first line gets its prefix from the template; subsequent directives carry their own prefix.

That matches the pre-2026-02-23 template, but [ironic-image#918](https://github.com/metal3-io/ironic-image/pull/918) (merged 2026-02-23) changed it to:

```jinja
{%- for item in env.DHCP_RANGE.split(";") %}
dhcp-range={{ item }}
{%- endfor %}
```

With the current template, a `\n`-separated `DHCP_RANGE` has no `;` separators, so `split(";")` returns one element — the entire multi-line blob — which gets substituted into a single `dhcp-range={{ item }}` expansion. That happens to produce valid `dnsmasq.conf` lines because each subsequent line is already a complete directive, but it's fragile, and it leads to issue #2.

### 2. Per-range gateways are emitted twice

`buildDHCPRange` interleaves `dhcp-option=tag:<n>,option:router,<gw>` lines into `DHCP_RANGE`, and `buildDHCPOptions` puts the same options into `DHCP_OPTIONS` (which #998's template loop also expands as `dhcp-option=` lines). The "ranges with tags and gateways interleaved" test case will render in `dnsmasq.conf` as:

```
dhcp-range=set:mgmt,10.0.0.10,10.0.0.100,255.255.255.0
dhcp-option=tag:mgmt,option:router,10.0.0.1        # from DHCP_RANGE blob
dhcp-range=set:pxe,192.168.1.10,192.168.1.200,255.255.255.0
dhcp-option=tag:pxe,option:router,192.168.1.1      # from DHCP_RANGE blob
dhcp-option=tag:mgmt,option:router,10.0.0.1        # from DHCP_OPTIONS (#998)
dhcp-option=tag:pxe,option:router,192.168.1.1      # from DHCP_OPTIONS (#998)
```

Each gateway gets a duplicated `dhcp-option=` line. dnsmasq probably copes, but this is redundancy by accident rather than by design.

### 3. Suggested simplification: align on `;` separation throughout

The cleanest fix is to drop the embedded-prefix machinery and use `;`-separated, prefix-free items in both env vars — matching the pattern that's already established for `DHCP_HOSTS` and now `DHCP_OPTIONS`:

- `DHCP_RANGE=set:mgmt,10.0.0.10,10.0.0.100,255.255.255.0;set:pxe,192.168.1.10,192.168.1.200,255.255.255.0`
- `DHCP_OPTIONS=tag:mgmt,option:router,10.0.0.1;tag:pxe,option:router,192.168.1.1`

This renders cleanly through the existing template (since #918) and through #998's loop, with no asymmetric first-line behavior, no embedded directive prefixes, and no duplication. `buildDHCPRange` becomes "join the per-range bodies with `;`" and the per-range options live exclusively in `buildDHCPOptions`.

### 4. dtantsur's version-gate concern (above) is still open

`newDnsmasqContainer` always emits `DHCP_OPTIONS` regardless of `versionInfo`. Presumably we want to reject `ranges` (or just reject `ranges` with `GatewayAddress`?) when the ironic-image version is older than the release that ships #918 + #998. I don't see that check anywhere in this PR yet.

### 5. Relaxed `provIP ∈ CIDR` check is broader than the relay use case

`ValidateDHCP` skips the `provCIDR.Contains(provIP)` check whenever `len(ranges) > 0`, even when the primary flat range is also configured. A user who sets both `networkCIDR`/`rangeBegin`/`rangeEnd` *and* a `ranges` entry can have a primary range whose CIDR doesn't contain the provisioning IP, and the validator stays silent. The primary range check is probably still meaningful even with `ranges` set — only the cross-list "must be in some CIDR" relationship needs to relax for the relay case.

### 6. Smaller notes

- **Field name**: `ranges` is fine; #553 used `networkRanges` for symmetry with the sibling `networkCIDR` fields — I'm easy either way.
- **Flat-field deprecation**: #553 marked the top-level `networkCIDR`/`rangeBegin`/`rangeEnd` as deprecated. This PR keeps them first-class. Two valid ways to express the same single range is API debt, but I understand the non-breaking call.
- **Auto-tag**: `range_<i+1>` only applies when a single range is configured without a `Name` (multi-range requires `Name`). Minor: when a user adds a second range later, validation forces them to name the first, so the tag changes from `range_1` to whatever they pick. A short doc comment on the field would be enough; not worth a code change.

Happy to be wrong on any of this — particularly the template-behavior trace. Will close #553 once this lands.

ftarasenko:
@holmboe Thanks for the careful read — you were right on all five substantive points. To my bad, I verified this deeply as an operator, not as a developer: confirmed it worked end-to-end + tests passed, and didn't dig deeper into the wire-format/template interaction. Pushed update to address them:
  1. **Wire format** — `buildDHCPRange` now returns a `;`-separated, prefix-free list. No more embedded `dhcp-range=` on subsequent items. New `TestDHCPRangePrefixFree` locks the invariant.
  2. **Duplicate routers** — `buildDHCPOptions` is now the sole source of `dhcp-option=…option:router` content. `TestNoDuplicateRouterOption` guards against regression.
  3. **`;` throughout** — done as part of (1)/(2). DHCP_RANGE, DHCP_OPTIONS, and DHCP_HOSTS now share the same pattern.
  4. **Version gate** — added `versionMultiRangeDHCP = Version360` and a `CheckVersion` arm that rejects `Ranges` on older Ironic. Covers both #918 and #998 (assumes they ship together in 36.0 — noted in a comment, easy to split if #998 slips).
  5. **provIP ∈ CIDR** — the containment check now fires whenever the flat range is configured. Relay relaxation kicks in only when no flat range is set. New test case covers the typo'd-flat-CIDR scenario.

  Plus three things I noticed while in there:
  - Per-range `gatewayAddress` must now live inside its `NetworkCIDR` (caught silently-broken configs).
  - Added `DHCPRange.MTU` (option:mtu emitted via DHCP_OPTIONS) — and dropped the matching field from our downstream gotmpl, which was previously generating CRD-invalid YAML.
  - `GATEWAY_IP` is zeroed in relay-only mode so the flat fallback router doesn't leak.

  Added a doc-comment on `Name` covering the auto-tag rename hazard you flagged. Kept `ranges` (not `networkRanges`) and the flat fields first-class — both per your "easy either way" / non-breaking call.

  Verified end-to-end on our k0s cluster: pre-fix `dnsmasq.conf` had 14 `option:router` lines for 7 ranges (each duplicated, exactly as you predicted); post-fix has 7, `dnsmasq --test` clean.

api/v1alpha1/ironic_types.go
```diff
@@ -30,6 +30,7 @@ const (
 
 var (
 	VersionLatest = Version{}
+	Version360    = Version{Major: 36, Minor: 0}
```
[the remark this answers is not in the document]
dtantsur: On top of this, I think the next version will be 37.0 because we skipped 36.0. Right @elfosardo?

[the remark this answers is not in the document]
dtantsur: @ftarasenko you can work around this problem by checking for `> 35.0` instead of `>= 37.0`.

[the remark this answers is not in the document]
ftarasenko: No problem, I'll update the code.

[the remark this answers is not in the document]
ftarasenko: Fixed

api/v1alpha1/ironic_types.go
```diff
@@ -60,13 +60,40 @@ type Inspection struct {
 	VLANInterfaces []string `json:"vlanInterfaces,omitempty"`
 }
 
+// DHCPRange defines a single DHCP address range with per-range options.
+type DHCPRange struct {
+	// Name is used as a dnsmasq tag for per-range options. Required when
+	// multiple ranges are defined; auto-generated as `range_<i+1>` otherwise.
```
dtantsur: This is confusing: what is `i` in `range_<i+1>` if it only applies to a single range?

api/v1alpha1/ironic_types.go
```diff
@@ -82,15 +109,27 @@ type DHCP struct {
...
 	RangeBegin string `json:"rangeBegin,omitempty"`
 
 	// RangeEnd is the last IP that can be given to hosts. Must be inside NetworkCIDR.
+	//
+	// Deprecated: use Ranges instead.
 	RangeEnd string `json:"rangeEnd,omitempty"`
 
+	// Ranges is a list of DHCP address ranges for multi-subnet support.
+	// Can be combined with the top-level flat range. When only Ranges is set,
+	// the provisioning IP need not be in any range's CIDR (DHCP relay).
```
dtantsur: If we deprecate top-level options, then we should expect only Ranges to be set even for cases other than DHCP relay. Maybe enforce that provisioningIP is in the first range in this case?

pkg/ironic/validation.go
```diff
@@ -121,53 +126,121 @@ func validateAgentImages(images []metal3api.AgentImages) error {
...
 		return err
 	}
 
-	if err := validateIP(dhcp.GatewayAddress); err != nil {
-		return err
+	hasPrimaryRange := dhcp.NetworkCIDR != "" || dhcp.RangeBegin != "" || dhcp.RangeEnd != ""
+	hasRanges := len(dhcp.Ranges) > 0
+
+	if !hasPrimaryRange && !hasRanges {
+		return errors.New("networking.dhcp.networkCIDR is required when DHCP is used")
```
dtantsur: Since you're deprecating networkCIDR, this should say "at least one range in networking.dhcp.ranges is ..."

pkg/ironic/version.go
```diff
@@ -125,5 +130,9 @@ func CheckVersion(resources Resources, version metal3api.Version) error {
 		return errors.New("networking service is only supported in Ironic 35.0 or newer")
 	}
 
+	if dhcp := resources.Ironic.Spec.Networking.DHCP; dhcp != nil && len(dhcp.Ranges) > 0 && version.Compare(versionMultiRangeDHCP) <= 0 {
+		return errors.New("networking.dhcp.ranges requires Ironic newer than 35.0")
```
dtantsur: You cannot do this if you deprecate the top-level options: there is no values that a user can set for 35.0 and older without triggering a deprecation warnings.

We need to either un-deprecate top-level options and rename "ranges" to "extraRanges" or make ranges with a single element work for older versions.

api/v1alpha1/ironic_types.go
ftarasenko: might next change be ok?

        // Name is used as a dnsmasq tag for per-range options. It must be set
        // (and unique) when more than one range is defined. When left empty, it
        // defaults to `range_N`, where N is the range's 1-based position in the
        // list. Must match [A-Za-z0-9_.-]+.

dtantsur: > position in the list

Which list are you talking about here? There is only one range per

> It must be set (and unique) when more than one range is defined

dtantsur:
We've discussed this issue with @ftarasenko and agreed to revert the deprecation of the top-level properties. Instead, we'll call the new field `ExtraRanges` to highlight their purpose. This resolves not just my comment about backward compatibility, but also the Copilot's concern about IPv6 support.

api/v1alpha1/ironic_types.go
dtantsur: Per our discussion: drop the Name field, always auto-generate tags for now.

holmboe:
Sounds good! Will be excellent to have this merged. 👏

pkg/ironic/version.go
ftarasenko: Old code. Now we do not deprecate top-level options.

api/v1alpha1/ironic_types.go
ftarasenko: Old code. Now we do not deprecate top-level options.

pkg/ironic/validation.go
ftarasenko: Old code. Now we do not deprecate top-level options.

api/v1alpha1/ironic_types.go
ftarasenko: Added rangeTag that returns the auto-generated dnsmasq tag for every extra range item

ftarasenko:
- Deprecation of the top-level properties is reverted: `networkCIDR`/`rangeBegin`/`rangeEnd`/`gatewayAddress` stay first-class and required whenever DHCP is enabled. Existing CRs work unchanged.
- The new field is `extraRanges` — strictly additional ranges on top of the main one (e.g. subnets behind a DHCP relay). The relay-only special cases are gone; the dnsmasq probe now covers mixed-family setups (port 67 and/or 547 depending on the configured CIDRs).
- Per your comment on `Name`: dropped, tags are always auto-generated as `range_N` (1-based position in `extraRanges`).
- The main `gatewayAddress` no longer leaks into extra ranges: it renders as the untagged `option:router`, so extra ranges without their own gateway get a valueless `tag:range_N,option:router` to suppress it (a main-subnet router is useless in any other IPv4 subnet).
- While in there, tightened validation: reversed ranges, /0 CIDRs and overlapping pools are rejected at admission instead of crash-looping dnsmasq.

One follow-up: some network designs must not serve DHCP on the provisioning subnet at all (an existing DHCP server owns it, or provisioning traffic only arrives via relays), while the main `networkCIDR` is still meaningful — it is where the provisioning IP lives. I'd like to add an optional
`serveMainRange` (defaults to true, so fully backward compatible): when false, only `extraRanges` are rendered into DHCP_RANGE, `networkCIDR` stays required (it still anchors the provisioning-IP check and probe selection), and `rangeBegin`/`rangeEnd` become optional. I have this implemented and
tested on a branch — would you prefer it in this PR or as a follow-up?

dtantsur:
> serveMainRange (defaults to true, so fully backward compatible): when false, only extraRanges are rendered into DHCP_RANGE, networkCIDR stays required (it still anchors the provisioning-IP check and probe selection), and rangeBegin/rangeEnd become optional.

Let's discuss it separately. I'd consider not adding the serveMainRange option and only making rangeBegin/rangeEnd optional with the same semantics.

api/v1alpha1/ironic_types.go
```diff
@@ -62,13 +60,42 @@ type Inspection struct {
...
 type DHCP struct {
 	// DNSAddress is the IP address of the DNS server to pass to hosts via DHCP.
 	// Must not be set together with ServeDNS.
 	// +optional
 	DNSAddress string `json:"dnsAddress,omitempty"`
 
+	// ExtraRanges is a list of additional DHCP address ranges served on top of
+	// the main range, e.g. for subnets reached via a DHCP relay. The main
+	// networkCIDR/rangeBegin/rangeEnd fields are still required. Requires
+	// Ironic newer than 35.0.
```
dtantsur: nit: at this point you can say "37.0 or newer)

pkg/ironic/version.go
```diff
@@ -20,6 +20,11 @@ var (
 
 	versionBMCCA      = metal3api.Version330
 	versionNetworking = metal3api.Version350
+
+	// versionMultiRangeDHCP gates ExtraRanges: split DHCP_RANGE and
+	// DHCP_OPTIONS support lands after 35.0; the next version number is not
+	// known yet, so check "newer than" instead of an exact minimum.
```
dtantsur: We know it now, it's 37.0

pkg/ironic/containers.go
```diff
@@ -638,13 +638,76 @@ func buildIronicHttpdPorts(ironic *metal3api.Ironic) (ironicPorts []corev1.Conta
 	return ironicPorts, httpdPorts
 }
 
+func prefixToNetmask(prefix netip.Prefix) string {
```
dtantsur: nit: could be moved to utils.go to avoid bloating this file

pkg/ironic/version.go
ftarasenko: @dtantsur, this change has already been merged into 37.0. Should I bump the version here as well?

pkg/ironic/validation.go
```diff
@@ -142,35 +204,76 @@ func ValidateDHCP(ironic *metal3api.IronicSpec) error {
...
+		addPool(r.RangeBegin, r.RangeEnd, fmt.Sprintf("networking.dhcp.extraRanges[%d]", i))
+	}
+
+	for i := 1; i < len(pools); i++ {
+		for j := range i {
+			if pools[i].begin.Compare(pools[j].end) <= 0 && pools[j].begin.Compare(pools[i].end) <= 0 {
+				return fmt.Errorf("DHCP pools of %s and %s overlap", pools[j].field, pools[i].field)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
ftarasenko: Copilot is wrong. range over an integer has been valid Go since 1.22, and this module is on go 1.25.0. It iterates j from 0 to i-1, exactly as intended. It compiles, go vet passes, and golangci-lint reports 0 issues (its intrange linter actually prefers this form). I left the code as-is. Copilot's claim reflects pre-1.22 knowledge.

pkg/ironic/containers.go
ftarasenko: Moved prefixToNetmask to utils.go with a doc comment; removed it from containers.go.

pkg/ironic/containers.go
```diff
@@ -638,13 +638,66 @@ func buildIronicHttpdPorts(ironic *metal3api.Ironic) (ironicPorts []corev1.Conta
...
-		return "" // don't disable your webhooks people
+	var parts []string
+
+	if dhcp.NetworkCIDR != "" && dhcp.RangeBegin != "" && dhcp.RangeEnd != "" {
+		prefix, err := netip.ParsePrefix(dhcp.NetworkCIDR)
+		if err == nil {
+			if len(dhcp.ExtraRanges) > 0 {
+				parts = append(parts, fmt.Sprintf("%s,%s,%s", dhcp.RangeBegin, dhcp.RangeEnd, prefixToNetmask(prefix)))
+			} else {
+				return fmt.Sprintf("%s,%s,%d", dhcp.RangeBegin, dhcp.RangeEnd, prefix.Bits())
```
dtantsur: I'm a bit confused about the difference here. Cannot you use the same code in both cases?

dtantsur:
/lgtm

Thanks!

pkg/ironic/containers.go
ftarasenko: dnsmasq parses dhcp-range fields by content, not position: it recognizes a netmask only when the token is a dotted IPv4 address, and an unadorned integer is taken as the lease time. So the old else branch emitting 192.168.1.10,192.168.1.200,24 would have set a 24-second lease with no netmask, rather than a /24 — a real bug, not just stylistic. IPv6 correctly uses the bare prefix length there, which prefixToNetmask preserves.

i've updated code and tests, but this bug was not introduced here, so please help me to check and confirm that fix is ok.

api/v1alpha1/ironic_types.go
elfosardo, line 33: can you please reabse? we have version370 already on main

ftarasenko: done

dtantsur:
/lgtm

Thanks! CI will be fixed by https://github.com/metal3-io/ironic-standalone-operator/pull/760

dtantsur:
Could you rebase to pick up the CI fixes?

ftarasenko:
> Could you rebase to pick up the CI fixes?

@dtantsur done

PR #674, opened 2026-05-15
:seedling: Bump golang to v1.25.10 and golang.org/x/net to v0.53.0

kashifest:
The bump is to fix several CVEs


lentzi90 (commented):
/approve

PR #679, opened 2026-05-18
✨ Callback and image-server url settings in Ironic image

1nv0k32:

In this PR, we are creating CRD fields to set `IRONIC_EXTERNAL_CALLBACK_URL` and `IRONIC_EXTERNAL_HTTP_URL` variables in Ironic image.
Also added `DisableHostNetwork` to disable ironic pod's hostnetworking.

Fixes #

**Checklist:**


1nv0k32:
Need to rebase after #655

dtantsur:
Would it be possible to un-tangle this PR from the Ingress one or do they conflict too much?

1nv0k32:
> Would it be possible to un-tangle this PR from the Ingress one or do they conflict too much?

Needs to be rebased after #655. I need the ingress part cause it needs to validate settings of this options with ingress:
https://github.com/metal3-io/ironic-standalone-operator/pull/679/changes#diff-ccca4f1e969c7dddd98e5aed466dbe2cb07413258a0945b71ad1bde5a56d2b43R196
As you mentioned before, setting either of these settings alone can cause unexpected behaviour in Ironic

1nv0k32:
I removed the pieces from #655 so the review would be easier but it doesn't build cause it depends on that PR. Should be merged after that :D

api/v1alpha1/ironic_types.go
```diff
@@ -160,6 +160,11 @@ type Networking struct {
 	// +optional
 	APIPort int32 `json:"apiPort,omitempty"`
 
+	// HostNetwork makes Ironic pod use the host network, which is required for Ironic to be accessed by external machines using the host IP address.
+	// This option would be considered as true unless the user explicitly sets it to false or sets ingress configuration, which implies HostNetwork=false.
```
dtantsur: This is a bit confusingly worded, I'm afraid.

First, to simplify the default logic, let's call it `disableHostNetworking`, to make it more obvious that the default value is `true`.

Second, let's make sure that it can only be set to `true` at all if either `externalIP` or `ingress` is configured.

Then you have something like

```go
// DisableHostNetwork turns off the HostNetwork flag on the Ironic pod, making it listen on the pod network.
// This option can only be used if `externalIP` or `ingress` are configured, since otherwise Ironic API and Image Server won't be accessible for hosts being provisioned.
// The default value is `true` if `ingress` is configured and `false` otherwise.
```

or along these lines

api/v1alpha1/ironic_types.go
```diff
@@ -177,10 +182,28 @@ type Networking struct {
 
 	// Configure Ingress resource for Ironic services.
 	// Set this option when you are planning to deploy Ironic in a public cluster and willing to use Ingress instead of IP address and NodePort.
+	// The operator should use this in case of virtual media deployments. The API and the image server will be accessed via the hostname specified in the ingress configuration.
```
dtantsur: > The operator should use this in case of virtual media deployments. 

This sounds like virtual media deployments won't work without this, which is not true. I think you meant something like "Ingress can only be used for virtual media deployments".

... which makes me reflect on the previous comment: if configuring Ingress turns HostNetwork to false, non-virtual media deployments won't be possible even if you configure a provisioning network. You need to clarify this aspect. Maybe we leave `DisableHostNetwork` as `false` by default and make it clear that setting it to `true` makes network boot impossible?

api/v1alpha1/ironic_types.go
```diff
@@ -177,10 +182,28 @@ type Networking struct {
...
 	// +optional
 	Ingress *Ingress `json:"ingress,omitempty"`
 
+	// externalCallbackURL for Ironic API server.
+	// Set this option when your Ironic API server is not directly accessible.
+	// Setting this option, will override URL set by networking.ingress.host.
+	// Must be set together with networking.imageServerExternalURL or networking.ingress
+	// +kubebuilder:validation:Format=uri
+	// +optional
+	ExternalCallbackURL string `json:"externalCallbackURL,omitempty"`
```
dtantsur, line 191: I'm not sure how I feel about these options because without Ingress a user needs to understand Ironic ports and with Ingress you already have a Host. What's the use case here that Host does not address?

pkg/ironic/containers.go
```diff
@@ -94,14 +94,19 @@ func buildCommonEnvVars(ironic *metal3api.Ironic) []corev1.EnvVar {
 		)
 		networkingProvided = true
 	}
+	// When networking.ingress is set or hostNetwork is disabled, we can provide the podIP
+	fieldPath := "status.hostIP"
```
dtantsur: nit: move the addition under `if !networkingProvided`

api/v1alpha1/ironic_types.go
dtantsur: These options also conflict with ExternalIP, this needs to be resolved if we add them.

pkg/ironic/validation.go
```diff
@@ -199,6 +199,12 @@ func ValidateIronic(ironic *metal3api.IronicSpec, old *metal3api.IronicSpec) err
 		return errors.New("networking.ingress and networking.externalIP cannot be set at the same time")
 	}
 
+	if ironic.Networking.Ingress == nil &&
+		((ironic.Networking.ImageServerExternalURL != "" && ironic.Networking.ExternalCallbackURL == "") ||
+			(ironic.Networking.ImageServerExternalURL == "" && ironic.Networking.ExternalCallbackURL != "")) {
+		return errors.New("in case of networking.ingress is not enabled, networking.imageServerExternalURL and networking.externalCallbackURL must be set together")
```
dtantsur: These options must always be set together regardless of Ingress. Setting one without the other is a footgun I'd rather avoid.

1nv0k32: That's why this validation is here...
Ingress will set both of them at the same time, then you can override either of them if you wish.
If Ingress is not set, the validation will check if both of the options are set together.

dtantsur: > then you can override either of them if you wish.

We should not allow exactly this. It nearly never makes sense.

1nv0k32: But why? Ingress sets both of fields to the ingress address by default, but consider the scenario that your OOB network is isolated and can't reach out to the image server and you need to proxy its call, then you have the opportunity to only override the `ImageServerExternalURL` (What we are currently doing in our deployment)

api/v1alpha1/ironic_types.go
1nv0k32: Ingress can operate if hostNetwork is true or false.
Network boot will fail if hostNetwork is false, so I put validation to check if host networking is disabled with all the stuff needed for network boot.

api/v1alpha1/ironic_types.go
lentzi90: You mean we should take the Ingress Host as external callback URL? That sounds reasonable to me.

lentzi90: Alternatively, the ingress host should be set from this?

lentzi90: Sorry, I should read before commenting... So this is the override for the ingress host. I am not sure what the use case is but I can imagine wanting to expose ironic in "some other way". For example through gateway API?

api/v1alpha1/ironic_types.go
```diff
@@ -160,13 +160,21 @@ type Networking struct {
 	// +optional
 	APIPort int32 `json:"apiPort,omitempty"`
 
+	// DisableHostNetwork disables the use of host networking for Ironic pods.
+	// Disabling host networking causes network boot impossible.
+	// kubebuilder:default=false
+	// +optional
+	DisableHostNetwork bool `json:"disableHostNetwork,omitempty"`
```
lentzi90: How about using positive polarity (`EnableHostNetwork` or `HostNetworkEnabled`) with default value of `true`? We get less double negatives that way. "Was host network not disabled?" => "Was host network enabled?"

pkg/ironic/validation.go
lentzi90: Yeah I think this can make sense. Also, users may want to expose it "manually" instead of relying on the ingress.

lentzi90 (commented):
This is mostly LGTM. I added an API design question below. Should we try to stick to positive polarity? I don't have a hard opinion on this, just want to lift it now while there is time to still change it.

api/v1alpha1/ironic_types.go
1nv0k32: As you also commented below, sometimes you don't want the Ingress host to be set in Ironic as callback/image url (e.g network is isolated and you need proxy to reach out to the ingress endpoint), So you use these two fields to override.

api/v1alpha1/ironic_types.go
1nv0k32: But `DisableHostNetwork` would signal that HostNetwork is enabled by default and you don't need to do anything about it unless you would like to disable it (like you are doing redfish virtual media and deploying metal3 on a multi tenant cluster, so you need to disable host networking for security reason).

lentzi90: Yeah I know it can be confusing in this way. There seem to be different philosophies about how to do it. Some want to go with the "default value principal", like you do here. Then the field should be named according to what the default value is and the default should be the true default of the type AND the default value for the field. E.g. we have host networking by default, the default value of a bool is false => field must be negative.

The other way of doing things is to stick with positive polarity. Then you don't necessarily use the default value of the type. Instead you set the default explicitly. Here that would mean something like EnableHostNetwork with default value `true`.

Like mentioned, I don't have a strong preference but I wanted to lift it in case other maintainers have.

lentzi90 (commented):
/label tide/merge-method-squash
/lgtm
@dtantsur please note the discussion about disable/enable host network. As mentioned I don't have a strong opinion on which way we go here so I am interested in what you think.

api/v1alpha1/ironic_types.go
```diff
@@ -160,13 +160,21 @@ type Networking struct {
 	// +optional
 	APIPort int32 `json:"apiPort,omitempty"`
 
+	// EnableHostNetwork enables the use of host networking for Ironic pods.
+	// Disabling host networking makes network boot impossible.
+	// +kubebuilder:default=true
+	// +optional
+	EnableHostNetwork *bool `json:"enableHostNetwork,omitempty"`
```
dtantsur: I remember asking but then missed the answer, so could you remind me: why not `DisableHostNetwork` and avoid a pointer altogether? I think earlier iterations provided different defaults for this value depending on Ingress, but it does not seem to be the case any more, right?

api/v1alpha1/ironic_types.go
```diff
@@ -176,11 +184,31 @@ type Networking struct {
...
 	Ingress *Ingress `json:"ingress,omitempty"`
 
+	// externalCallbackURL for Ironic API server.
+	// Set this option when your Ironic API server is not directly accessible.
+	// Setting this option, will override URL set by networking.ingress.host.
+	// Must be set together with networking.imageServerExternalURL or networking.ingress
+	// Cannot be set at the same time with networking.externalIP.
+	// +kubebuilder:validation:Format=uri
+	// +optional
+	ExternalCallbackURL string `json:"externalCallbackURL,omitempty"`
```
dtantsur, line 191: nit: could you keep the fields sorted?

api/v1alpha1/ironic_types.go
```diff
@@ -176,11 +184,31 @@ type Networking struct {
...
+	// Ingress should only be used with virtual media deployments.
 	// Cannot be set at the same time with networking.externalIP.
 	// +optional
 	Ingress *Ingress `json:"ingress,omitempty"`
 
+	// externalCallbackURL for Ironic API server.
+	// Set this option when your Ironic API server is not directly accessible.
+	// Setting this option, will override URL set by networking.ingress.host.
+	// Must be set together with networking.imageServerExternalURL or networking.ingress
+	// Cannot be set at the same time with networking.externalIP.
```
dtantsur, line 188: Please copy the comment about virtual media from the ExternalIP docstring (same for image)

api/v1alpha1/ironic_types.go
dtantsur: Hmm, I ended up leaving a comment about not doing it :-/ `EnableHostNetwork` sounds like it's disabled by default, which is not the case.

dtantsur: So far, IrSO has been using `Disable*` fields for things that are enabled by default (e.g. DisableDownloader). I'd rather not change it.

lentzi90: No hard feelings! I am fine with `Disable*`, especially if we have that for other fields already :slightly_smiling_face: 
I only brought it up because we ended up going with positive polarity throughout in CAPO

lentzi90:
We should mark this as feature I think.
/retitle ✨ Callback and image-server url settings in Ironic image

api/v1alpha1/ironic_types.go
1nv0k32: Back to Disable again :D

api/v1alpha1/ironic_types.go
1nv0k32: I moved it to DisableHostNetwork and removed the pointers, seems cleaner now.
Regarding the different defaults, I ditched that plan. IMO, that feels obscure to me to change the value of hostnetwork based on other fields like Ingress.
Better to have the host networking enabled and give the operator the option to disable it for better security in case they are only doing virtual media and enforcing it if they are doing network boot.

pkg/ironic/validation.go
```diff
@@ -187,6 +187,11 @@ func ValidateIronic(ironic *metal3api.IronicSpec, old *metal3api.IronicSpec) err
 		return errors.New("credentialsName, host and name are required on database")
 	}
 
+	if ironic.Networking.DisableHostNetwork &&
+		(ironic.Networking.BindInterface || ironic.Networking.DHCP != nil || ironic.Networking.Interface != "" || ironic.Networking.IPAddress != "" || len(ironic.Networking.MACAddresses) > 0 || ironic.Networking.Keepalived != nil) {
+		return errors.New("networking.enableHostNetwork cannot be set to false together with networking.bindInterface or networking.dhcp or networking.interface or networking.ipAddress or networking.macAddresses or networking.keepalived")
```
dtantsur: Old field name here (must be `disableHostNetwork`)

dtantsur:
/approve
/copilot-review

Please fix one issue (and whatever copilot detects)

pkg/ironic/containers.go
```diff
@@ -410,10 +414,25 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
...
+			imageServerExternalURL = resources.Ironic.Spec.Networking.ImageServerExternalURL
+		}
+	} else if resources.Ironic.Spec.Networking.ExternalCallbackURL != "" && resources.Ironic.Spec.Networking.ImageServerExternalURL != "" {
+		// only set the external URLs when both are provided, to avoid misconfiguration
+		externalCallbackURL = resources.Ironic.Spec.Networking.ExternalCallbackURL
+		imageServerExternalURL = resources.Ironic.Spec.Networking.ImageServerExternalURL
+	}
+	if externalCallbackURL != "" && imageServerExternalURL != "" {
+		result = appendStringEnv(result, "IRONIC_EXTERNAL_CALLBACK_URL", externalCallbackURL)
+		result = appendStringEnv(result, "IRONIC_EXTERNAL_HTTP_URL", imageServerExternalURL)
```
Rozzii, line 435: I think there is a use-case that has been missed here.
I might be wrong here when reading the code and you might be handling the use-case already, but I would like to highlight a scenario where an external image server is in use even if HostNetworking is enabled. There are environments where users are virtualmedia booting with pre-built IPA ISOs and the ISOs are downloaded from an external server. This use-case is usually relevant in  resource constrained edge computing environments. 

For reasons mentioned above,  I suggest to allow setting imageServerExternalURL by itself but if externalCallbackURL is set then that would make imageServerExternalURL required.

Rozzii (commented):
In general the code is very nice a very verbose, feature(s) make sense, but I have one issue with it.

pkg/ironic/containers.go
dtantsur: This option is about the HTTP server from the Ironic pod, not some arbitrary server. Setting this option makes ironic assume that any virtual media images from its local cache are available via that URL.

1nv0k32: @dtantsur @Rozzii 
Do we have final decision on this matter?
I kinda agree with @dtantsur comment as this option is for the whole http server and ironic relies on it even for the images it makes.
Maybe for having different http servers for different purposes we can introduce another feature? :thinking:

dtantsur: Also keep in mind: it's much easier to relax a restriction later than to introduce one.

PR #685, opened 2026-05-21
[WIP] 🐛 Do not set Ironic variables for trusted ca if tls is not configured

### LINKED ISSUE #684: Settings trustedCA without certificateName causes Ironic to setup TLS
https://github.com/metal3-io/ironic-image/blob/main/scripts/tls-common.sh#L70

If you set `ironic.spec.tls.trustedCA` , ironic pod will be created with `IRONIC_CACERT_FILE` which causes `IRONIC_TLS_SETUP` to become true which causes below config to be enabled:

https://github.com/metal3-io/ironic-image/blob/main/ironic-config/ironic.conf.j2#L268

Which is not desired in case you just want to trust a CA but not enable ssl on ironic.

https://github.com/metal3-io/ironic-standalone-operator/commit/2eab9ab280796600856540f836975ef31290202f
Probably this can be the cause.

1nv0k32:

Fixes #684

**Checklist:**

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -268,16 +268,20 @@ func buildTrustedCAEnvVars(cctx ControllerContext, resources Resources) []corev1
 	// Build the path to the CA bundle file
 	caPath := fmt.Sprintf("%s/ca/trusted/%s", certsDir, selectedKey)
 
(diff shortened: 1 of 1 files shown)


1nv0k32:
Closing in favor of #689

PR #686, opened 2026-05-22
:sparkles: Add configurable Keepalived VRID and VRRP password support

ftarasenko:

  Extend the Keepalived sidecar configuration with two new optional fields under `Networking`:

  - **`keepalivedVRID`** (`int32`) — configures the VRRP virtual router ID. Default behaviour is unchanged (uses the value baked into the upstream keepalived image template). When set to a non-default value, the
   operator injects a command override that patches `keepalived.conf` before the daemon starts. The sed expression is anchored and uses a digit character class, so any VRID value in the template (1, 10, 51, …)
  is rewritten correctly. This avoids VRID conflicts when multiple Ironic instances share the same L2 broadcast domain.

  - **`keepalivedPasswordRef`** (`*ResourceReference`) — references a Secret carrying a VRRP simple-auth password. The password is exposed to the keepalived container as the `KEEPALIVED_AUTH_PASS` env var
  sourced from the referenced Secret's `password` key. Because the upstream keepalived image template ships *without* an authentication block, the operator inserts one (`auth_type PASS` + `auth_pass
  ${KEEPALIVED_AUTH_PASS}`) before the `virtual_ipaddress` line at runtime.

  **Cross-field validation**: setting `keepalivedVRID > 1` or `keepalivedPasswordRef` now requires `networking.ipAddressManager` to be `keepalived`. This replaces the previous silent no-op behaviour when those
  fields were set without the corresponding manager.

  Tests cover env-var wiring, sed-script content for VRID/password combinations, the auth-block-insertion shape, and the new validation cases.

  Fixes #

  **Checklist:**


dtantsur:
/ok-to-test

Let's wait for https://github.com/metal3-io/ironic-standalone-operator/pull/636/changes because it will add a separate `Keepalived` structure that you can extend.

P.S.
@ftarasenko if you want to receive any responses to your private messages, I highly recommend not making messaging you cost Telegram money ;) I didn't even know it was possible...

ftarasenko:
> /ok-to-test
> 
> Let's wait for https://github.com/metal3-io/ironic-standalone-operator/pull/636/changes because it will add a separate `Keepalived` structure that you can extend.
> 
> P.S. @ftarasenko if you want to receive any responses to your private messages, I highly recommend not making messaging you cost Telegram money ;) I didn't even know it was possible...

Sorry for that. Fixed. Rebased cause 636 landed.

api/v1alpha1/ironic_types.go
```diff
@@ -216,6 +216,24 @@ type Networking struct {
...
 	RPCPort int32 `json:"rpcPort,omitempty"`
+
+	// KeepalivedVRID is the VRRP virtual router ID used by keepalived.
+	// Must be unique within the L2 broadcast domain.
+	// Only applies when IPAddressManager is "keepalived".
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=255
+	// +optional
+	KeepalivedVRID int32 `json:"keepalivedVRID,omitempty"`
```
dtantsur: Please move the new options into KeepalivedConfig

api/v1alpha1/ironic_types.go
```diff
@@ -134,6 +134,22 @@ type KeepalivedConfig struct {
...
+	// +optional
+	VRID int32 `json:"vrid,omitempty"`
+
+	// PasswordRef references a Secret with the VRRP simple-auth
+	// (PASS) password under the key "password". VRRP truncates the value to
+	// 8 bytes — longer values are silently cut by keepalived.
+	// The password must consist of characters safe to embed unquoted in a
+	// keepalived.conf line (printable ASCII, no whitespace, no '"' or '\').
+	// +optional
+	PasswordRef *ResourceReference `json:"passwordRef,omitempty"`
```
lentzi90, line 152: I think we should have an explicit Secret reference here, not a ResourceReference that can be either Secret or ConfigMap.

pkg/ironic/containers.go
```diff
@@ -776,7 +776,27 @@ func newKeepalivedContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) c
...
+			envVars = append(envVars, corev1.EnvVar{
+				Name: "KEEPALIVED_AUTH_PASS",
+				ValueFrom: &corev1.EnvVarSource{
+					SecretKeyRef: &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{Name: pw.Name},
+						Key:                  "password",
+					},
+				},
+			})
+		}
```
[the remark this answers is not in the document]
lentzi90: We should not allow ConfigMap here.

pkg/ironic/containers.go
```diff
@@ -791,6 +811,35 @@ func newKeepalivedContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) c
...
+			// inserted lines don't contain that string, so the address only
+			// matches the original line on every pass and the block lands in
+			// the correct order.
+			script.WriteString("sed -i '/virtual_ipaddress/i\\    authentication {' /etc/keepalived/keepalived.conf\n")
+			script.WriteString("sed -i '/virtual_ipaddress/i\\        auth_type PASS' /etc/keepalived/keepalived.conf\n")
+			script.WriteString("sed -i \"/virtual_ipaddress/i\\        auth_pass ${KEEPALIVED_AUTH_PASS}\" /etc/keepalived/keepalived.conf\n")
+			script.WriteString("sed -i '/virtual_ipaddress/i\\    }' /etc/keepalived/keepalived.conf\n")
+		}
+		script.WriteString("exec /bin/manage-keepalived.sh\n")
+		container.Command = []string{"/bin/bash", "-c", script.String()}
```
lentzi90, line 839: This feels very brittle. Why not add support for these first in the keepalived image? We would get much nicer and less code here. I do not think IRSO should be the entity handling the logic around keepalived anyway. That fits better in the image itself.

dtantsur: I concur. Unless we decide to move all scripts into IrSO (which is a topic for another day), it's better to keep all logic in the image. Since keepalived image does not have releases, you'll be able to use the new change right away.

pkg/ironic/containers.go
```diff
@@ -791,6 +811,35 @@ func newKeepalivedContainer(versionInfo VersionInfo, ironic *metal3api.Ironic) c
...
+	// configured. The upstream keepalived image ships a template that has
+	// virtual_router_id hardcoded and no authentication block at all, so:
+	//   - VRID: sed-replace the existing value (anchored, digit class).
+	//   - Password: insert an authentication { auth_type PASS; auth_pass ... }
+	//     block before the virtual_ipaddress line. The password value is
+	//     sourced from a Secret-backed env var at runtime.
+	if vrid > 1 || hasPassword {
+		var script strings.Builder
+		if vrid > 1 {
+			fmt.Fprintf(&script, "sed -i -E 's/^([[:space:]]*)virtual_router_id [0-9]+/\\1virtual_router_id %d/' /etc/keepalived/keepalived.conf\n", vrid)
```
Rozzii, line 825: I very much don't like this, I would rather change https://github.com/metal3-io/utility-images/blob/main/keepalived/sample.keepalived.conf to use a template variable so that we don't need to use sed here.

pkg/ironic/containers.go
Rozzii: > I concur. Unless we decide to move all scripts into IrSO (which is a topic for another day), 

I don't think we have to include anything other than ironic-image pod parameter/env var configuration in the scope of IrSO. The ability to run ironic-image on its own is an advantage and not a disadvantage IMO. 


> it's better to keep all logic in the image

Strongly agree, also it is much more maintainable and simpler to just change the template https://github.com/metal3-io/utility-images/blob/main/keepalived/sample.keepalived.conf then using sed in this controller.

Rozzii (commented):
Thanks for the contribution, this feature is valid and needed.
I have issues with this PR though, I don't support the use of "sed" in this context, I wrote down my opinion below.

Are you planning to continue the work @ftarasenko? You have been absent for some time and the PR is at risk of getting stale.

ftarasenko:
@Rozzii I’m planning to review your comments carefully, complete the remaining work, and thoroughly test the changes. I’ll need some time to do this properly, so please keep the pull request open.

api/v1alpha1/ironic_types.go
ftarasenko: Done here.

ftarasenko:
@Rozzii can you check https://github.com/metal3-io/utility-images/pull/80 for the template changes and approve the names? If everything looks good, I’ll update this PR accordingly

Rozzii:
> @Rozzii can you check [metal3-io/utility-images#80](https://github.com/metal3-io/utility-images/pull/80) for the template changes and approve the names? If everything looks good, I’ll update this PR accordingly

Sure I will check that.

PR #689, opened 2026-05-27
🐛 Fix IRONIC_CACERT_FILE triggering TLS without a certificate

### LINKED ISSUE #684: Settings trustedCA without certificateName causes Ironic to setup TLS

alegacy:
Setting spec.tls.trustedCA without spec.tls.certificateName caused ironic-image to enable TLS because IRONIC_CACERT_FILE was always set from trustedCA. The trustedCA is meant for verifying external services (WEBSERVER_CACERT_FILE), not for Ironic's own TLS endpoints.

Decouple IRONIC_CACERT_FILE from trustedCA and only set it when TLS is enabled (certificateName is configured):

1. If the TLS secret contains ca.crt (e.g. from cert-manager), use it
2. Otherwise, fall back to the trustedCA path
3. If neither is available, don't set it (ironic-image assumes self-signed and copies the leaf cert as the CA)

Without TLS, IRONIC_CACERT_FILE is never set regardless of trustedCA.

Fixes: #684

**Checklist:**

- [X] Unit tests have been added, if necessary.
- [X] E2E tests have been added, if necessary.

### DIFF
--- a/pkg/ironic/containers.go
+++ b/pkg/ironic/containers.go
@@ -270,11 +270,7 @@ func buildTrustedCAEnvVars(cctx ControllerContext, resources Resources) []corev1
 
 	return []corev1.EnvVar{
 		{
-			Name:  "WEBSERVER_CACERT_FILE", // CA for verifying image server TLS connections
-			Value: caPath,
-		},
-		{
-			Name:  "IRONIC_CACERT_FILE", // CA for verifying JSON-RPC TLS connections (e.g. ironic-networking)
+			Name:  "WEBSERVER_CACERT_FILE",
 			Value: caPath,
 		},
 	}
@@ -380,7 +376,32 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
 	}
 
 	if resources.TrustedCAConfigMap != nil || resources.TrustedCASecret != nil {
-		result = append(result, buildTrustedCAEnvVars(cctx, resources)...)
+		trustedCAEnvVars := buildTrustedCAEnvVars(cctx, resources)
+		result = append(result, trustedCAEnvVars...)
+
+		// When TLS is enabled, set IRONIC_CACERT_FILE for JSON-RPC TLS verification.
+		// Prefer the CA from the TLS secret (ca.crt, populated by cert-manager);
+		// fall back to the trusted CA bundle if the TLS secret lacks ca.crt.
+		if resources.TLSSecret != nil {
+			if _, hasCACert := resources.TLSSecret.Data["ca.crt"]; hasCACert {
+				result = append(result, corev1.EnvVar{
+					Name:  "IRONIC_CACERT_FILE",
+					Value: certsDir + "/ironic/ca.crt",
+				})
+			} else if len(trustedCAEnvVars) > 0 {
+				result = append(result, corev1.EnvVar{
+					Name:  "IRONIC_CACERT_FILE",
+					Value: trustedCAEnvVars[0].Value,
+				})
+			}
+		}
+	} else if resources.TLSSecret != nil {
+		if _, hasCACert := resources.TLSSecret.Data["ca.crt"]; hasCACert {
+			result = append(result, corev1.EnvVar{
+				Name:  "IRONIC_CACERT_FILE",
+				Value: certsDir + "/ironic/ca.crt",
+			})
+		}
 	}
 
 	if resources.Ironic.Spec.ExtraConfig != nil {
--- a/pkg/ironic/containers_test.go
+++ b/pkg/ironic/containers_test.go
@@ -390,9 +390,9 @@ func TestTrustedCAConfigMap(t *testing.T) {
 			}
 			assert.Equal(t, expectTrustedCA, foundMount, "Volume mount existence mismatch")
 
-			// Check environment variables (WEBSERVER_CACERT_FILE and IRONIC_CACERT_FILE)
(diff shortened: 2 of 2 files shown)


alegacy:
## Support Matrix: `IRONIC_CACERT_FILE` behavior after this fix                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                  
  **IrSO logic (this PR):**                                                                                                                                                                                                                                                                                       
  - `WEBSERVER_CACERT_FILE`: always set from `trustedCA` when configured (unchanged)                                                                                                                                                                                                                              
  - `IRONIC_CACERT_FILE`: set when TLS is configured                                                                                                                                                                                                                              
    1. Prefer `ca.crt` from the TLS secret (e.g. populated by cert-manager)                                                                                                                                                                                                                                       
    2. Fall back to the `trustedCA` path                                                                                                                                                                                                                                                                          
  - Without TLS, `IRONIC_CACERT_FILE` is **never** set regardless of `trustedCA`                                                                                                                                                                                                                                  

  **ironic-image fallback** (when `IRONIC_CACERT_FILE` is not set by IrSO):
  - If TLS cert exists but no CA file: copies the leaf cert as the CA (assumes self-signed)

  | # | TLS | CA type | trustedCA | `ca.crt` in TLS secret? | `IRONIC_CACERT_FILE` | `IRONIC_TLS_SETUP` | Result |
  |---|-----|---------|-----------|:-:|---|:-:|---|
  | 1 | No | — | No | — | not set | false | works |
  | 2 | No | — | Yes | — | not set | false | works **(#684 fixed)** |
  | 3 | Yes | Self-signed | No | Maybe | not set (leaf fallback) | true | works (leaf=CA) |
  | 4 | Yes | Self-signed | Yes | Yes | `/certs/ironic/ca.crt` | true | works |
  | 5 | Yes | Self-signed | Yes | No | trustedCA path | true | works |
  | 6 | Yes | Private CA (cert-mgr) | No | Yes | not set (leaf fallback) | true | works |
  | 7 | Yes | Private CA (manual) | No | No | not set (leaf fallback) | true | User must configure `trustedCA` |
  | 8 | Yes | Private CA (cert-mgr) | Yes | Yes | `/certs/ironic/ca.crt` | true | works |
  | 9 | Yes | Private CA (manual) | Yes | No | trustedCA path | true | works |
  | 10 | Yes | Public CA (ACME) | No | No | not set (leaf fallback) | true | works (chain in `tls.crt`) |

  **Notes:**
  - Scenarios 7 requires the user to configure `trustedCA` with the issuing CA. This is expected — a private CA is not in the system trust store, so it must be explicitly provided.
  - Scenario 10 works because ACME issuers typically include the full chain (leaf + intermediate) in `tls.crt`, and ironic-image copies the entire file as the CA.

diconico07:
I think we should be able to cover Scenario 6 without the need for the user to do anything if we set `IRONIC_CACERT_FILE` to the `ca.crt` field if TLS is enabled but `trustedCA` is unset.

Scenario 5 and 9 are requiring the `trustedCA` secret/cm to actually contain the needed certificate, while I don't have anything against this, I believe it should be made quite explicit in some documentation.

alegacy:
> I think we should be able to cover Scenario 6 without the need for the user to do anything if we set `IRONIC_CACERT_FILE` to the `ca.crt` field if TLS is enabled but `trustedCA` is unset.

Right, I've updated the code and retroactively updated the support matrix above to reflect that change.

> Scenario 5 and 9 are requiring the `trustedCA` secret/cm to actually contain the needed certificate, while I don't have anything against this, I believe it should be made quite explicit in some documentation.

Ya, 5 is a bit tricky while I expect that 9 would probably be true in most cases.

lentzi90 (commented):
Copilot comments looks like nits to me.
/lgtm

alegacy:
Any other feedback here?

pkg/ironic/containers.go
```diff
@@ -380,7 +376,32 @@ func buildIronicEnvVars(cctx ControllerContext, resources Resources) []corev1.En
 	}
 
 	if resources.TrustedCAConfigMap != nil || resources.TrustedCASecret != nil {
-		result = append(result, buildTrustedCAEnvVars(cctx, resources)...)
+		trustedCAEnvVars := buildTrustedCAEnvVars(cctx, resources)
+		result = append(result, trustedCAEnvVars...)
+
+		// When TLS is enabled, set IRONIC_CACERT_FILE for JSON-RPC TLS verification.
+		// Prefer the CA from the TLS secret (ca.crt, populated by cert-manager);
+		// fall back to the trusted CA bundle if the TLS secret lacks ca.crt.
```
dtantsur, line 384: Could you document this behavior somewhere on the API level, so that it's user-visible?

dtantsur:
/approve

Please address the nits in a follow-up when you have a minute

PR #693, opened 2026-06-01
🐛 Fix dnsmasq probe for IPv6-only deployments

### LINKED ISSUE #692: dnsmasq probes hardcoded for DHCPv4, breaks IPv6-only deployments
Version: v0.9.0  
Environment: IPv6-only provisioning network

## Problem

The dnsmasq container's liveness and readiness probes are hardcoded to check for port 67 (DHCPv4):

`pkg/ironic/containers.go` (~line 520):
```
probe := newProbe(corev1.ProbeHandler{
    Exec: &corev1.ExecAction{
        Command: []string{"sh", "-c", "ss -lun | grep :67 && ss -lun | grep :69"},
    },
})
```

In an IPv6-only environment, dnsmasq does not bind to port 67. It binds to:
- Port 547 (DHCPv6)
- Port 69 (TFTP)

This causes the readiness/liveness probes to fail indefinitely, putting the dnsmasq container into CrashLoopBackOff and preventing the Ironic CR from reaching Ready state.

## Impact

- Ironic CR stays in DeploymentInProgress / NotReady
- Bare metal provisioning cannot proceed because DHCP/TFTP services are considered unhealthy
- The only workarounds are post-deployment kubectl patch or experimental spec.overrides.containers replacement

## Reproduction
1. Deploy IrSO v0.9.0
2. Create an Ironic CR with IPv6-only networking:
     ```
     networking:
       ipAddress: fd00:abcd:ef01:3fff::a
       dhcp:
         networkCIDR: fd00:abcd:ef01:3fff::/64
         rangeBegin: fd00:abcd:ef01:3fff::3000
         rangeEnd: fd00:abcd:ef01:3fff::3fff
     ```
3. Observe the ironic-service pod — dnsmasq container fails probes and restarts continuously

## Current Workarounds

| Approach | Drawback |
|----------|----------|
| kubectl patch `deployment/ironic-service` after deployment | Not declarative; breaks GitOps |
| `spec.overrides.containers` to replace entire dnsmasq container | EXPERIMENTAL; requires redefining full container spec; brittle on upgrades |

## Suggested Fix

Option A (Preferred): Add `dnsmasqLivenessProbe` and `dnsmasqReadinessProbe` fields to `spec.overrides`, matching the existing `httpdLivenessProbe` / `httpdReadinessProbe` pattern.

Option B: Make the default dnsmasq probe IPv6-aware — detect if networkCIDR is IPv6 and check port 547 instead of 67.

holmboe:
## Summary

> **UPDATE:** Revised per review feedback. The dnsmasq probe override
> fields have been removed so this PR is a pure, backportable bugfix.
> The original commit was signed off (DCO), and the override removal is
> a separate follow-up commit so the history stays clear. The probe
> override escape hatch can be proposed separately later.

- The dnsmasq container's liveness/readiness probe was hardcoded to check
  UDP `:67` (DHCPv4) and `:69` (TFTP). On IPv6-only provisioning networks
  dnsmasq binds to `:547` (DHCPv6) instead of `:67`, so the probe failed
  forever and the Ironic CR never reached `Ready`.
- The probe now picks the DHCP port from the parsed
  `networking.dhcp.networkCIDR` (`547` for IPv6, `67` for IPv4); TFTP `:69`
  is unchanged.

Fixes #692

## Test plan

- [x] `go test ./pkg/ironic/... -run TestDnsmasqProbeConfiguration` — 2 scenarios pass (IPv4 → `:67`, IPv6 → `:547`).
- [x] `TestExpectedContainers` and `TestHttpdProbeConfiguration` still pass.
- [x] `make generate manifests` regenerated `zz_generated.deepcopy.go`, the CRD YAML, and `docs/api.md`.
- [x] Functional verification on an IPv6-only K3s cluster: apply an Ironic CR with `networking.dhcp.networkCIDR: fd00:abcd:ef01:3fff::/64` and confirm dnsmasq reaches `Ready` instead of `CrashLoopBackOff`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dtantsur:
/ok-to-test

Could you please propose the new overrides for probes separately? I'm not sure we even need them, and we definitely cannot have them in a bugfix since it will prevent us from backporting it.

Please also sign-off your commit per https://github.com/metal3-io/ironic-standalone-operator/pull/693/checks?check_run_id=78874624125

holmboe:
Thanks for the review @dtantsur! Updated accordingly:

- **Dropped the probe overrides** — removed the `DnsmasqLivenessProbe` /
  `DnsmasqReadinessProbe` fields so this is now a pure bugfix that can be
  backported. I'll propose the override escape hatch separately if it
  turns out to be needed.
- **Signed off the commits** (DCO). I kept the original commit (now
  signed off) and added a separate follow-up commit for the override
  removal so the history stays clear in this PR.

The IPv6-only port selection (547 for DHCPv6, 67 for DHCPv4) is
unchanged.

PR #698, opened 2026-06-04
Add timeouts for all workflows

### LINKED ISSUE #667: Set timeout for short github workflows
Sometimes a github workflow might get stuck, e.g. due to a random network hickup or some temporary issue with a mirror.
When this happens, it would be nice to cancel the workflow within a reasonable amount of time. The default timeout is 360 minutes which is a ridiculously long time to wait for a linter.

We should set suitable timeouts for the workflows, especially those that are expected to run only for a short time.

Check a recent PR for what workflows are triggered and how long they run.

Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepstimeout-minutes

UncivilizedGhost:
**What this PR does / why we need it**: Github workflows might get stuck and it takes 6 hours to cancel by default 
Timeouts were set as follows:

- 5 minutes for all the smaller workflows (that are 30 seconds or less on average) 
- 10 minutes for medium workflows (Average around 2 max around 5) 
- 15 for one workflow which reached up to 8 minutes
- 90 for the largest workflows (most average around 20 but the highest is can reach 90 minutes)

Fixes https://github.com/metal3-io/ironic-standalone-operator/issues/667

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -14,6 +14,7 @@ on:
 jobs:
   build_controller:
     name: Build Ironic-standalone-operator container image
+    timeout-minutes: 10
     if: github.repository == 'metal3-io/ironic-standalone-operator'
     permissions:
       contents: read
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: read
+    timeout-minutes: 10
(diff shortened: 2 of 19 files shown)


lentzi90 (commented):
Thanks! This looks good, beside the copilot comment. Please fix that, then I think it is good to go.
/retest

lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #720, opened 2026-06-18
🌱 Use proper IPA cache address

Rozzii:
This commit:
 - Makes sure that the correct IPA nordix proxy is used.

Previous address circumvented the proper cache refresh trigger so in cas


tuminoid:
/retitle 🌱 Use proper IPA cache address

Not product bug.

tuminoid (commented):
/approve

lentzi90 (commented):
/lgtm

PR #723, opened 2026-06-18
🌱 Set container image workflow job timeout to 10 min

migiFi:
Set container image workflow job timeout to 10 min

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -141,6 +141,7 @@ jobs:
       id-token: write
     needs: push_release_tags
     name: Build IrSO container image
+    timeout-minutes: 10
     if: github.repository == 'metal3-io/ironic-standalone-operator'
     uses: metal3-io/project-infra/.github/workflows/container-image-build.yml@main # zizmor: ignore[unpinned-uses]
     with:


Rozzii (commented):
/approve

cmd/main.go
```diff
@@ -143,6 +143,10 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
+	if databaseImage != "" {
+		setupLog.Info("warning: mariadb-image is deprecated. Use MariaDB Operator instead: https://github.com/mariadb-operator/mariadb-operator")
```
dtantsur: "deprecated" means it's still present but should not be used. It actually does not work at all, so maybe turn it into an error?

migiFi: thank you, I have changed to fail with `os.Exit(1)` instead of just logging a warning.

dtantsur:
Could you decouple this change from the release workflow change?

migiFi:
> Could you decouple this change from the release workflow change?

 [#759](https://github.com/metal3-io/ironic-standalone-operator/pull/759)

dtantsur:
It looks like you've left the wrong change here, could you double-check?

Rozzii:
Now this PR is nothing more than 1 timeout change, please check @migiFi

migiFi:
> Now this PR is nothing more than 1 timeout change, please check @migiFi

@Rozzii  seems like the  bumps were by the dependabot [here ](https://github.com/metal3-io/ironic-standalone-operator/commit/52f6117ce7934ef04b86b8530bcbf450b4c4bcb5#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05ada) 

so there is no more changes to make here besides the timeout witch is not on the main, but is it needed at all?

dtantsur:
The subject says

> Deprecate mariadb-image flag

have we actually done it?

Rozzii:
> The subject says
> 
> > Deprecate mariadb-image flag
> 
> have we actually done it?

Yes, I have found it here https://github.com/metal3-io/ironic-standalone-operator/pull/759
So part of this work wos done by https://github.com/metal3-io/ironic-standalone-operator/pull/759 and the other part by dependabot https://github.com/metal3-io/ironic-standalone-operator/commit/52f6117ce7934ef04b86b8530bcbf450b4c4bcb5#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05ada
I will rename the PR and change the PR message, @migiFi please change the commit message.
Also I am not familiar with the reason why the timeout is introduced so if it is not relevant anymore please close the PR.

migiFi:
seems like this is not needed anymore, the changes here were propose by @UncivilizedGhost  but reverted by @tuminoid  here [#736 ](https://github.com/metal3-io/ironic-standalone-operator/pull/736/changes) so this change is not relevant.

PR #734, opened 2026-06-24
🌱Add fuzzing test for ironic standalone operator

smoshiur1237:
Fuzz test for the ironic-standalone-operator using Go's built-in fuzzing support. 
We are going to add initial setting for fuzz test which will validate the ironic using `FuzzValidateIronic` function and tried to add both valid and invalid corpus seed. 

The `FuzzValidateIronic` will detects panics for arbitrary IP addresses, CIDR ranges, DHCP ranges, and image URLs. Validation errors are expected. The only invariant is that no panic occurs.

This will be an initial fuzz test and possibility to add more fuzzing option in future.


smoshiur1237 (commented):
/cc @tuminoid @Rozzii

test/fuzz/README.md
```diff
@@ -0,0 +1,38 @@
+# Fuzzing
+
+Fuzz tests for the ironic-standalone-operator using Go's built-in fuzzing support.
+
+## Running Fuzz Tests
+
+### Quick Start with Makefile
+
```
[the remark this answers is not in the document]
smoshiur1237: This test will be used from workflow. I don't think it is needed at this point.

dtantsur:
Are you plan to running it in the CI? Otherwise, it's easy to regress.

Makefile
```diff
@@ -144,6 +144,22 @@ test: manifests generate fmt vet ## Run tests.
 	go test ./... -coverprofile cover.out
 	cd api && go test ./... -coverprofile cover.out
 
+FUZZ_TIME ?= 30s
+
+.PHONY: fuzz
+fuzz: ## Run fuzz tests with seed corpus (no fuzzing, regression test only)
+	cd $(TEST_DIR)/fuzz && go test -race -v ./...
```
lentzi90, line 151: This could maybe be moved under the test target. Then `fuzz-run` could be renamed to just `fuzz`.

lentzi90 (commented):
/lgtm
One nit below

smoshiur1237:
> Are you plan to running it in the CI? Otherwise, it's easy to regress.

Yes, I will add a reusable workflow which will run on weekly basis. Same as BMO workflow: [Fuzz workflow](https://github.com/metal3-io/baremetal-operator/blob/main/.github/workflows/scheduled-fuzz.yml)

This will be an initial fuzzing test, gradually we can add more test and update the seeds according to our needs.

Makefile
smoshiur1237: This was intentional to do `fuzz `tests as regression tests (seed corpus only, fast) and `fuzz-run` for proper fuzzing with thousands of valid and invalid samples of seeds.  But if we want, can do improvement later.. The same approach was using in other repo such as BMO, IPAM and CAPM3 but we can improve it separately.

Rozzii (commented):
/approve

PR #735, opened 2026-06-25
:seedling: add 0.10 to dependabot, drop 0.8 and 0.7

tuminoid:
Per version support 0.10 and 0.9 are supported and 0.8 tested, 0.7 EOL.


lentzi90 (commented):
/lgtm

PR #736, opened 2026-06-25
:seedling: workflows reusing other workflows cannot have timeouts

tuminoid:
Its a syntax error. The timeout must be on the reused workflow side.

All reusing workflows in IRSO, including release.yaml are broken.

### DIFF
--- a/.github/workflows/build-images-action.yml
+++ b/.github/workflows/build-images-action.yml
@@ -14,7 +14,6 @@ on:
 jobs:
   build_controller:
     name: Build Ironic-standalone-operator container image
-    timeout-minutes: 10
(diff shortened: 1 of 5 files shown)


tuminoid:
/hold

Some of them are real, but let me debug a bit more before blanket removing them all.

tuminoid:
/unhold

They really are broken, and not even visible in PR jobs. Which means we have merged plenty of PRs without functional job. Luckily we have periodic functional passing, so nothing should be broken.

/cc @elfosardo @dtantsur @lentzi90 @Rozzii

lentzi90 (commented):
/approve

PR #741, opened 2026-06-25
:rocket: Republish 0.10.0 release notes to trigger release creation

elfosardo:


tuminoid (commented):
/approve
/lgtm

Lets see if this fixes things or we need to republish as v0.10.1

PR #747, opened 2026-07-01
🌱 Remove zizmor inline ignore for release-0.10

migiFi:
Removed the # zizmor: ignore[artipacked] inline suppression comment from .github/workflows/release.yaml. The artipacked warning is now handled via a targeted ignore rule in .zizmor.yml at the repo root instead, since persist-credentials: false cannot be used here as credentials are needed later in the workflow.


Rozzii (commented):
/approve

PR #758, opened 2026-07-15
🌱 Bump go.opentelemetry.io/otel/sdk to 1.43.0 and x/net to 0.55

UncivilizedGhost:
**What this PR does / why we need it**:Fix the following CVEs

 - https://osv.dev/GHSA-hfvc-g4fc-pqhx 
 - https://osv.dev/GO-2026-4918
 - https://osv.dev/GO-2026-5026

Fixes #

**Checklist:**


api/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/metal3-io/ironic-standalone-operator/api
 
-go 1.24.0
+go 1.25.0
```
dtantsur, line 3: This needs an impact analysis. Bumping Go has a good chance of breaking downstreams and must be communicated with a ⚠️ in the release notes. But before we need to decide if the issues are critical enough to accept the potential breakage.

/cc @Rozzii @tuminoid

Rozzii: Will take a look, I went through recent OSV scanner findings for the  CAPM3, BMO, and IPAM patch releases and there were quite a lot of false positives or "non exploitable"  vulns so this might be similar.

PR #759, opened 2026-07-15
 🌱 Deprecate mariadb-image flag

migiFi:
Deprecate mariadb-image flag by converting it to a fatal error.

The flag no longer works and users should migrate to MariaDB Operator instead. This change fails early with an error message directing users to the migration path: https://github.com/mariadb-operator/mariadb-operator


Rozzii (commented):
/approve

PR #760, opened 2026-07-22
:seedling: CI: regenerate manifests and fix broken link checker

dtantsur:

### DIFF
--- a/.lycheeignore
+++ b/.lycheeignore
@@ -1,0 +1,2 @@
+# Kubernetes moved these examples away. They end up in docs/api.md, which is auto-generated.
+https://examples.k8s.io/(mysql-cinder-pd|volumes)/.*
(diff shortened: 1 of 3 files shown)


.github/workflows/pr-link-check.yml
```diff
@@ -1,14 +0,0 @@
-name: PR Check Links
-
-on:
-  pull_request:
-    types: [opened, edited, reopened, synchronize, ready_for_review]
```
Rozzii, line 5: Instead of removing try 
paths-ignore:
      - 'docs/api.md'

paths-ignore goes under "types"

Rozzii (commented):
I think I have an idea
/hold

.github/workflows/pr-link-check.yml
Rozzii: Okay after further discussion and looking into the issue just create a .lycheeignore file and andd the 3 problematic links to the file please, and open an issue so someone can fix it later.

Rozzii (commented):
/lgtm

PR #769, opened 2026-07-30
🌱 Harden hack script against WORKDIR volume mount injection

migiFi:
Affects the following scripts under the hack/ directory:

- shellcheck.sh

This script read WORKDIR from the env and insert it directly into the
container's --volume mount specification. Because the volume spec is colon-delimited
(`host:target:options`), a WORKDIR value containing a colon could alter how the mount
is parsed.

This change removes the WORKDIR var and hardcodes the container path to /workdir,
which matches the pattern already used in several other metal3 repos. Since /workdir
was already the default, there is no change in behavior for normal use, but the
injection surface is eliminated.

**NOTE**: the other scripts in this repo (gomod.sh, manifestlint.sh, and markdownlint.sh) already had hardcode /workdir. This change brings shellcheck.sh in line with them.

### DIFF
--- a/hack/shellcheck.sh
+++ b/hack/shellcheck.sh
@@ -5,7 +5,6 @@ set -eux
 
 IS_CONTAINER="${IS_CONTAINER:-false}"
 CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-podman}"
-WORKDIR="${WORKDIR:-/workdir}"
 
 if [ "${IS_CONTAINER}" != "false" ]; then
     TOP_DIR="${1:-.}"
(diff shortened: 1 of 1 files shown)


lentzi90 (commented):
The workdir changes looks good, but there are some unrelated changes included. Maybe started from another feature branch rather than `main`?
/hold

migiFi:
> Maybe started from another feature branch rather than `main`?

it did indeed. Its fixed now.

lentzi90 (commented):
/lgtm
/hold cancel
/retest

tuminoid (commented):
/approve

PR #772, opened 2026-08-03
🌱 Remove leftover MariaDB image code

migiFi:
Cleans up the unused MariaDBImage code that pointed at the old, now archive metal3-io/mariadb image. Nothing used this code anymore, so removing it doesnt chnage the behaviour.

The `--mariadb-image` flag still shows its "use MariaDB Operator instead" message.


Rozzii (commented):
/approve

PR #773, opened 2026-08-04
🌱 Dependabot cooldown fix and remove zizmor inline ignore

migiFi:
Change the dependabot cooldown from 7 days to 3 days. Instead of suppressing the zizmor warning with `zizmor: ignored`, this adds a proper `.zizmor.yml` config at the repo root. The workflow at `.github/workflows/zizmor.yml` picks it up via the `config:` input.

Removed the `# zizmor: ignore[artipacked]` inline suppression comment from `.github/workflows/release.yaml`. The `artipacked` warning is now handled via a targeted ignore rule in `.zizmor.yml` at the repo root instead, since `persist-credentials: false` cannot be used here as credentials are needed later in the workflow.

### DIFF
--- a/.zizmor.yml
+++ b/.zizmor.yml
@@ -1,0 +1,14 @@
+# zizmor configuration file
+# This file controls zizmor automation rules such as Dependabot throttling.
+# See zizmor documentation for more details:
+# https://docs.zizmor.sh/
+
+rules:
+  dependabot-cooldown:
+    config:
+      days: 3
+
+  # persist-credentials needed later in the workflow
+  artipacked:
+    ignore:
+    - release.yaml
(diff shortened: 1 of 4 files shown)


.zizmor.yml
```diff
@@ -0,0 +1,14 @@
...
+
+rules:
+  dependabot-cooldown:
+    config:
+      days: 3
+
+  # persist-credentials needed later in the workflow
+  artipacked:
+    ignore:
+    - release.yaml
```
lentzi90: nit: missing newline at end of file

lentzi90 (commented):
/lgtm

Rozzii (commented):
/approve

lentzi90 (commented):
/lgtm

PR #788, opened 2026-08-12
:warning: Drop support for Ironic 33.0 and 34.0

elfosardo:


pkg/ironic/version.go
```diff
@@ -17,7 +17,7 @@ var (
 	defaultRamdiskDownloaderImage = defaultRegistry + "/ironic-ipa-downloader:latest"
 	defaultKeepalivedImage        = defaultRegistry + "/keepalived:latest"
 
-	versionBMCCA      = metal3api.Version330
+	versionBMCCA      = metal3api.Version350
```
dtantsur: This is wrong: the version check needs to removed entirely

pkg/ironic/version.go
```diff
@@ -17,7 +17,7 @@ var (
...
 	versionNetworking = metal3api.Version350
```
dtantsur: nit: optionally remove this too (versions below 35 not possible)

lentzi90 (commented):
/lgtm

PR #814, opened 2026-08-20
:seedling: Local tests: retry podman commands if they hang 

dtantsur:
It happens under unclear conditions, and the resulting github actions
report is very unhelpful. Retry several times and let collection
continue.

Also don't silence the ps output: it can be helpful for debugging.


lentzi90 (commented):
/approve

PR #833, opened 2026-08-31
🐛 Do not set HostPorts when DisableHostNetwork is true

dtantsur:
Refresh existing unit tests to be more meaningful and also cover Ports.


lentzi90 (commented):
/approve

tuminoid (commented):
/lgtm

PR #843, opened 2026-09-08
:seedling: Add periodic jobs for release-0.10 and 0.11

### LINKED ISSUE #841: Add periodic jobs for release-0.10 and release-0.11
We should have periodic jobs like this: https://github.com/metal3-io/ironic-standalone-operator/blob/main/.github/workflows/functional-periodic-release-0.9.yml
for all supported release branches. Looks like we missed to update them, so they are now missing for 0.10 and 0.11. We could also clean up the older workflows. There is no need to test 0.8 and 0.7 anymore.
Reference for version support: https://book.metal3.io/version_support.html#ironic-standalone-operator

Abhayanthk:
Fixes #841

Adds periodic functional workflows for `release-0.10` and `release-0.11` (modeled on the existing `release-0.9` job), with cron schedules offset by a few minutes. Also removes the obsolete `release-0.7` and `release-0.8` periodic workflows, which are no longer supported per the version support policy.

### DIFF
--- a/.github/workflows/functional-periodic-release-0.10.yml
+++ b/.github/workflows/functional-periodic-release-0.10.yml
@@ -2,8 +2,8 @@ name: Periodic Functional Tests release-0.10
 
 on:
   schedule:
-  # Run every day at 03:25 UTC (it is recommended to avoid running at the start of the hour)
-  - cron: '25 3 * * *'
+  # Run every day at 05:20 UTC (it is recommended to avoid running at the start of the hour)
+  - cron: '20 5 * * *'
   workflow_dispatch:
 
 permissions: {}
--- a/.github/workflows/functional-periodic-release-0.10.yml
+++ b/.github/workflows/functional-periodic-release-0.10.yml
@@ -1,9 +1,9 @@
-name: Periodic Functional Tests release-0.8
+name: Periodic Functional Tests release-0.10
 
 on:
   schedule:
-  # Run every day at 04:20 UTC (it is recommended to avoid running at the start of the hour)
-  - cron: '20 4 * * *'
+  # Run every day at 03:25 UTC (it is recommended to avoid running at the start of the hour)
+  - cron: '25 3 * * *'
   workflow_dispatch:
 
 permissions: {}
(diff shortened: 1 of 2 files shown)


.github/workflows/functional-periodic-release-0.10.yml
```diff
@@ -1,9 +1,9 @@
-name: Periodic Functional Tests release-0.7
+name: Periodic Functional Tests release-0.10
 
 on:
   schedule:
-  # Run every day at 05:20 UTC (it is recommended to avoid running at the start of the hour)
-  - cron: '20 5 * * *'
+  # Run every day at 03:25 UTC (it is recommended to avoid running at the start of the hour)
+  - cron: '25 3 * * *'
```
lentzi90: Let's just keep this as it was. The jobs are nicely spaced already so the new job can run in the same slot the old job used.
```suggestion
  # Run every day at 05:20 UTC (it is recommended to avoid running at the start of the hour)
  - cron: '20 5 * * *'
```

.github/workflows/functional-periodic-release-0.11.yml
```diff
@@ -1,9 +1,9 @@
-name: Periodic Functional Tests release-0.8
+name: Periodic Functional Tests release-0.11
 
 on:
   schedule:
-  # Run every day at 04:20 UTC (it is recommended to avoid running at the start of the hour)
-  - cron: '20 4 * * *'
+  # Run every day at 03:30 UTC (it is recommended to avoid running at the start of the hour)
+  - cron: '30 3 * * *'
```
lentzi90: Here also we can keep it as it was.

```suggestion
  # Run every day at 04:20 UTC (it is recommended to avoid running at the start of the hour)
  - cron: '20 4 * * *'
```

lentzi90 (commented):
Thanks for the PR! Please keep the cron schedule as it was. Otherwise this is good.

Abhayanthk:
@lentzi90, you're absolutely right — I shouldn't have changed the original schedule times. That was my mistake. Thank you for the careful review and the correction; I've fixed it now. Please let me know if there's anything else I should address.

.github/workflows/functional-periodic-release-0.10.yml
Abhayanthk: Reverted, thanks!

lentzi90 (commented):
/approve
/label tide/merge-method-squash

PR #847, opened 2026-09-10
:seedling: Fix yamllint workflow

terror96:

The keyword 'timeout-minutes' is not allowed with 'uses' and causes a parsing error.


lentzi90 (commented):
Thank you!
/approve

tuminoid (commented):
/lgtm

PR #855, opened 2026-09-14
:seedling: Add id-token: write permission in build-images-action-pr workflow

kashifest:
Currently the workflow fails with 

```
Invalid workflow file: .github/workflows/build-images-action-pr.yml#L11
The workflow is not valid. .github/workflows/build-images-action-pr.yml (Line: 11, Col: 3): Error calling workflow 'metal3-io/project-infra/.github/workflows/container-image-build.yml@main'. The nested job 'job' is requesting 'id-token: write', but is only allowed 'id-token: none'.
```


Rozzii (commented):
/approve

lentzi90 (commented):
/lgtm

PR #872, opened 2026-09-19
:seedling: Run zizmor on release branches

### LINKED ISSUE #859: Configure zizmor to run on release branches
We initially added zizmor only to main. However, now that it the workflow and needed fixes also exist on release branches, we should update it to run also there.

See https://github.com/metal3-io/ironic-standalone-operator/blob/main/.github/workflows/zizmor.yml

nicknikolakakis:

Adds `release-*` to the `push` and `pull_request` branch filters in `.github/workflows/zizmor.yml`, matching the pattern already used in `build-images-action.yml`.

Workflows are resolved per ref, so a push to `release-0.11` or a PR targeting it uses that branch's copy of `zizmor.yml` rather than the one on `main`. This change therefore needs a cherry-pick to take effect on a release branch. I checked `release-0.9`, `release-0.10` and `release-0.11`: all three already carry `zizmor.yml` and `.zizmor.yml`, with the branch filter still set to `main` only, so a straight cherry-pick is enough for them. Happy to follow up with those once this lands, if you tell me how far back you want to go.

Unrelated to this change, but you will run into it the next time `zizmor-action` is bumped here: zizmor 1.30.0 added the `self-repository` audit, which flags five spots in this repo (`functional-pr.yml:11` and the four `functional-periodic-*.yml:14`). Not touched here to keep this PR to the trigger change.

Fixes #859

**Checklist:**

- [ ] Integration tests have been added, if necessary.


lentzi90 (commented):
Thanks!
/ok-to-test
/approve

Rozzii (commented):
/lgtm

PR #877, opened 2026-09-23
🌱Remove unused RBAC rule from IRSO

smoshiur1237:
The pods (get/list/watch) permission is not used anywhere. Every Pod reference in non-test code is a `PodTemplateSpec ` embedded inside `DaemonSet/Deployment` specs — the operator never reads, lists, or watches actual Pod objects, and `SetupWithManager` does not Owns/Watches Pods. This could be a leftover from the default kubebuilder scaffold.

That marker grants the operator read-only access to Pod objects cluster-wide, It grants no write access — the operator could not create, update, patch, or delete Pods with it. But in this code base it does nothing, no code actually performs any of these operations on Pod objects. The Pods that back the Ironic deployment are created and managed indirectly by the `DaemonSet/Deployment` controllers , so IRSO never touches Pod objects directly. That's why this marker was a leftover from the kubebuilder scaffold and was safe to remove.


smoshiur1237:
/cc @lentzi90 @dtantsur 
Would you please review and opinions? I think it is safe to remove the unused rule.

Rozzii (commented):
It was introduced intentionally here
https://github.com/metal3-io/ironic-standalone-operator/commit/0147c41cbeebe858264031f7393eaf9bf92641fc
3 years ago, but I am not sure why, for me it looks like there was an intention to do something with pods directly but it has never materialized, are we missing something @dtantsur  ?

lentzi90:
Hmm I could imagine that we check the pod status, but on the other hand, anything important should bubble up to higher level objects so that is probably enough.
/copilot-review

Rozzii (commented):
Neither us or AI can find any code that would explicitly touch Pods so I will approve.
/approve
We can put these rules back later if needed maybe eventually there will be some feature implemented that will need it.

tuminoid (commented):
/hold
Until more eyes.

tuminoid (commented):
LGTM. I don't see where these are needed, since IRSO doesn't actually watch for the Pods directly or via controller runtime, only DaemonSets themselves. But I still wouldn't be surprised if this would be some controller-runtime hidden requirement that shows up only in some use-case.

lentzi90 (commented):
/lgtm
also from my side

